PR #1, opened 2018-06-26
Copied Readme from docs

GunnarLieb:
Why not having a nicer Readme, I copied from the docs and changed the links


elpete:
Thanks!

PR #5, opened 2018-07-20
QUICK-17

wpdebruin:
enabled support for multidb and multi grammar

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -18,12 +18,13 @@ component accessors="true" {
     property name="mapping";
     property name="fullName";
     property name="table";
+    property name="queryoptions";
     property name="readonly"        default="false";
     property name="attributeCasing" default="none";
     property name="key"             default="id";
     property name="attributes";
     property name="meta";
-
+    
     /*=====================================
     =            Instance Data            =
     =====================================*/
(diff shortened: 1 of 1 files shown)


wpdebruin:
Just some remarks about what I did.

you can add datasource and grammar usermetadata for the component, example
```component table="domains" datasource="testMySQL" grammar="MySQLGrammar" extends="quick.models.BaseEntity"```

- new property ```queryoptions``` is set in ```MetaDataInspection()``` from metadata for datasource.
- datasource and grammar are only kept in metadata, not in properties. Do we need them?
- queryoptions is set to the datasource when nececessary, otherwise left empty.
- queryoption is added to all query and  persistence methods where necessary by looking for the following qb methods ```insert update delete get first find``` and looking for ```newquery()``` and ```getQuery()``` methods. 
-  following qb methods with queryoptions were not found in quick so not modified ```updateorInsert count max min sum aggegratequery exist value values implode run runQuery```

TODO
We should have a close look at the magic methods, to find out if they are passing query or persistence methods to qb without adding a datasource attribute
Testing: I tested some simple save actions with a mysql grammar while default grammar was mssql. And datasource was changed from the default. Obviously this needs more testing, but basics are here.

PR #6, opened 2018-07-27
Resolves QUICK-19

jclausen:
- Allows for memento-ization of inheritance chain
- Also allows `getMemento()` to return a full representation of the memento, even when variable is null ( e.g. - an unloaded entity )

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -3,35 +3,49 @@ component accessors="true" {
     /*====================================
     =            Dependencies            =
     ====================================*/
-    property name="builder"            inject="QueryBuilder@qb";
-    property name="wirebox"            inject="wirebox";
-    property name="str"                inject="Str@str";
-    property name="settings"           inject="coldbox:modulesettings:quick";
-    property name="validationManager"  inject="ValidationManager@cbvalidation";
-    property name="interceptorService" inject="coldbox:interceptorService";
-    property name="keyType"            inject="AutoIncrementing@quick";
+    property name="builder"
+                inject="QueryBuilder@qb" 
+                persistent="false";
+    property name="wirebox"
+                inject="wirebox" 
+                persistent="false";
+    property name="str"
+                inject="Str@str" 
+                persistent="false";
+    property name="settings"
+                inject="coldbox:modulesettings:quick" 
+                persistent="false";
+    property name="validationManager"
+                inject="ValidationManager@cbvalidation" 
+                persistent="false";
+    property name="interceptorService"
+                inject="coldbox:interceptorService"
+                persistent="false";
+    property name="keyType"
+                inject="AutoIncrementing@quick" 
+                persistent="false";
 
     /*===========================================
     =            Metadata Properties            =
     ===========================================*/
-    property name="entityName";
-    property name="mapping";
-    property name="fullName";
-    property name="table";
-    property name="readonly"        default="false";
-    property name="attributeCasing" default="none";
-    property name="key"             default="id";
-    property name="attributes";
-    property name="meta";
+    property name="entityName"      persistent="false";
+    property name="mapping"         persistent="false";
+    property name="fullName"        persistent="false";
+    property name="table"           persistent="false";
+    property name="readonly"        default="false" persistent="false";
+    property name="attributeCasing" default="none"  persistent="false";
+    property name="key"             default="id"    persistent="false";
+    property name="attributes"      persistent="false";
+    property name="meta"            persistent="false";
 
     /*=====================================
     =            Instance Data            =
     =====================================*/
-    property name="data";
-    property name="originalAttributes";
-    property name="relationshipsData";
-    property name="eagerLoad";
-    property name="loaded";
+    property name="data" persistent="false";
+    property name="originalAttributes" persistent="false";
+    property name="relationshipsData" persistent="false";
+    property name="eagerLoad" persistent="false";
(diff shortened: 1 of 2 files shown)


models/BaseEntity.cfc
```diff
@@ -761,15 +778,19 @@ component accessors="true" {
...
+
     private function setAttributesFromProperties( properties ) {
-        return setAttributes(
+
+        setAttributes(
             properties.reduce( function( acc, prop ) {
                 param prop.column = prop.name;
                 param prop.persistent = true;
-                if ( prop.persistent ) {
+                if ( javacast( "boolean", prop.persistent ) ) {
```
elpete: What cases are we covering with the `javacast`?

models/BaseEntity.cfc
```diff
@@ -749,7 +765,8 @@ component accessors="true" {
     }
 
     private function metadataInspection() {
-        var md = getMetadata( this );
+        var util = createObject("component","coldbox.system.core.util.Util");
```
elpete: Can we inject this instead of creating it inline?

models/BaseEntity.cfc
```diff
@@ -63,11 +77,13 @@ component accessors="true" {
     }
 
     function getAttributesData( aliased = false, withoutKey = false ) {
+
         getAttributes().keyArray().each( function( key ) {
-            if ( variables.keyExists( key ) && ! isReadOnlyAttribute( key ) ) {
-                setAttribute( key, variables[ key ] );
+            if ( ! isReadOnlyAttribute( key ) ) {
+              setAttribute( key, !isNull( variables[ key ] ) ? variables[ key ] : ''  );
```
elpete: I'm not sure I want to set `null` variables to empty strings.  I think I'd like to keep them null, especially with full null support on the horizon.

models/BaseEntity.cfc
jclausen: String values in property anotations if they don't specify a `type` attribute:

```
property name="myProp" persistent="true";
```

There are some inconsistencies with different engines on that, but javacast fixes them all.

models/BaseEntity.cfc
jclausen: I took my cues from how it's used in the framework. It's a one-off whenever it's used.  I can take a look at injection, though, since `BaseEntity` is a transient, it shouldn't be an issue.

models/BaseEntity.cfc
jclausen: That's fair, but then you're not going to have a representation of those keys in your memento in all existing engines ( and the ones with null support turned off ).  I would rather have an option to turn on full null support and default to having the key present.

elpete: That's been requested elsewhere, as well.  I think we need a module setting for this.  Probably outside the scope of this PR.

jclausen: At least, for now, I believe we should have a full representation of the memento.  Otherwise you need a default for every property, which may or may not represent the database return typing and value.

elpete: If full null was turned off, wouldn't the `queryExecute` call return an empty string anyway?

jclausen: After execution, yes. The problem is the state of the memento in a newly created entity.  There's no visibility in to the persistent columns through the memento or the attributes, because null is the same as it not existing.

You can say as a developer "I know this is a property", for of course, but this ensures that the key has visiblility.

models/BaseEntity.cfc
elpete: I think perhaps it can't be injected because WireBox hasn't injected anything yet — `init` vs. `onDIComplete`.

It really would be ideal if this were a separate package we could pull in as a dependency, but I'm not going to force that point.

models/BaseEntity.cfc
elpete: We should have visibility into the persisted columns because we have an array of the persistent attributes available in `getAttributes`.  Without `legacyNullSupport` it would return something like:

jclausen: The attributes are set from the metadata introspection.  They don't get set if the property is null, which is the reason for this pull request.  

When `getMemento()` returns `getAttributes` the null variables no longer exist.  What is `undefined`?  That's not a CF thing.  In an engine that doesn't have full null, you have no way to check for the available keys if it's a null because `structKeyArray()` and `for( var key in getMemento()` are going to exclude the nulls.

In any case, we shouldn't write a new module targeted at full null support, at this point. It's fine to be a bit opinionated and I wholly agree on trying to develop toward the future, but you have over 20+ years of null support and a half-dozen engine versions that have treated nulls as non-existent.  We need to develop with that in mind, IMHO.

elpete:
Can you check out these two commit?  I think it may solve the null issues you were facing.

https://github.com/coldbox-modules/quick/commit/3f27408412886f851a4da73a274c9223088ca2db
https://github.com/coldbox-modules/quick/commit/940db4dc2f2a0e423b81cd37d284be5206e5c55e

If so, can you rebase this PR off of development and we can get it merged?

jclausen:
@elpete : Please let me know if there are any blockers on getting this merged.  The inherited metadata is blocker for me.

elpete:
I think `development` has the inherited metadata fixes in it.  Did you check out the two commits listed above?

jclausen:
OK. I looked only at the commits and not at the dev branch `BaseEntity` in full.  It looks like that has been addressed.  I'll close this one out.

PR #7, opened 2018-08-08
QUICK-24: Add AssignedKey KeyType

MordantWastrel:
https://ortussolutions.atlassian.net/browse/QUICK-24

### DIFF
--- a/models/KeyTypes/AssignedKey.cfc
+++ b/models/KeyTypes/AssignedKey.cfc
@@ -1,0 +1,19 @@
+component implements="KeyType" {
+
+    /**
+     * Called to handle any tasks before inserting into the database.
+     * Recieves the entity as the only argument.
+     */
+    public void function preInsert( required entity ) {
+        return;
+    }
+
+    /**
+     * Called to handle any tasks after inserting into the database.
+     * Recieves the entity and the queryExecute result as arguments.
+     */
+    public void function postInsert( required entity, required struct result ) {
+        return;
+    }
+
+}


elpete:
Looks good.  Can you add a test for this, please?

MordantWastrel:
The current tests will not run on MSSQL:

If I create a datasource called 'quick' but don't create the DB, Lucee throws the error 'Cannot open database "quick" requested by the login. The login failed. '

If I create the datasource but don't create the DB, I get 'Cannot drop database "quick" because it is currently in use.'

elpete:
I should document this for running the tests, but please use a MySQL datasource.  Docker is a great option if you need one locally:

```
docker run -d --name=quick --env MYSQL_DATABASE=quick --env MYSQL_ROOT_PASSWORD=quick -p 3306:3306 mysql:5
```

MordantWastrel:
That's easy enough, but is there documentation somewhere on the proper procedure for testing coldbox modules? In the past it's been a guessing game of reading Application.cfc, figuring out where to put Coldbox, when to clone a github repo and when to use Forgebox. I know it isn't especially difficult from having done it in the past, but I don't remember offhand what I did and it would grease the wheels on opening this stuff up to have an authoritative guide for module tests (as there is for Coldbox).

elpete:
The right answer is it needs to be a part of the README.md for the repo.  That way anyone that clones the repo can run the tests.  I'll work on adding that as well as a `docker-compose.yml` file for making it as painless as possible.

elpete:
https://ortussolutions.atlassian.net/browse/QUICK-25

PR #8, opened 2018-08-24
refactor(BaseEntity): Make internal properties more resistant to name clashes

elpete:
This commit changes almost all the internal property names and functions to better
avoid name clashes.  This involved prefixing all properties with an underscore
as well as eliminating almost all public methods that start with `set` or `get`.

BREAKING CHANGE: Any code that used internal functions of the `BaseEntity`
will need to update to the name function and property names.

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -3,54 +3,53 @@ component accessors="true" {
     /*====================================
     =            Dependencies            =
     ====================================*/
-    property name="builder"            inject="QueryBuilder@qb";
-    property name="wirebox"            inject="wirebox";
-    property name="str"                inject="Str@str";
-    property name="settings"           inject="coldbox:modulesettings:quick";
-    property name="validationManager"  inject="ValidationManager@cbvalidation";
-    property name="interceptorService" inject="coldbox:interceptorService";
-    property name="keyType"            inject="AutoIncrementing@quick";
+    property name="_builder"            inject="QueryBuilder@qb";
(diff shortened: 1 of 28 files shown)


elpete:
@MordantWastrel @wpdebruin @jclausen 

I'd love to get your feedback on the above changes and how it might affect your development.

MordantWastrel:
This looks great!

I don't think it's necessary a problem for fields like **key** and **keyType** -- while it's conceivable that there could be overlap with user code on those variables, I was more concerned with the **getThing()** functions. I'd have no trouble reserving key and keyType for Quick, though.

wpdebruin:
I think it is a good idea, for me it makes things more clear.  I was not using quick (yet), so no problems here.

Tropicalista:
I don't know if this is related, however I post here because there's no bug tracking.

If I create a 

`	property name="settings" type="struct";
`

when I try to insert a record, 

```   
 function setSettings( value ) {
        return setAttribute( "settings", serializeJSON(value) );
 } 
```

I got this error:

Messages: can't assign value to a Object of this type [String] with key automaticValidation

Template: | F:\Progetti\CodeCanyon\Refactoring\apiservertest\modules\quick\models\BaseEntity.cfc
-- | --
LINE: | 929: param settings.automaticValidation = false;

This is probably a name issue, because if I try to change name, it works great.

elpete:
@Tropicalista That is exactly it.  That one has been changed in this PR as well.

PR #13, opened 2018-10-11
Fixes dependency chain for cfcollection

jclausen:


elpete:
`cfcollection` has been removed as a dependency of Quick.  It is now an optional dependency.  It is included in `devDependencies` so the test of `QuickCollection` still works.

jclausen:
If it's in the inheritance chain within the models, then it has to be a required dependency.  Otherwise module registration will fail when wirebox auto-maps the modules and inspects the metadata for singletons.

elpete:
Hey, @jclausen.  I moved `QuickCollection` to not be mapped by default, so this shouldn't be an issue anymore.  Thanks!

jclausen:
On `@be` it is still throwing an error due to the inheritance.  Is there a separate commit for this that isn't in the `development` branch?

```

Lucee 5.2.9.31 Error (expression)
--
Message | invalid component definition, can't find component [cfcollection.models.Collection]
Stacktrace | The Error Occurred in/Users/jonclausen/Sites/ortus/modules/cbCommerce/test-harness/modules/quick/models/QuickCollection.cfc: line 1 1: component extends="cfcollection.models.Collection" {2: 3: function collect( data ) {

```

jclausen:
Closing this out as it looks like the refactoring was in `#development` and not `@be`

PR #14, opened 2018-11-29
Fixes incorrect variable name

jclausen:
`belongsToMany` relationships may not be created until this PR is merged


elpete:
Can you resubmit against the new version of development?

PR #16, opened 2018-12-10
update qb dependency

joeltobey:
Updating to qb v 6. Had to make a small update due to this commit in qb:

qb commit 7b12b021ea3d96d382761ac8a931885430ced3d0
- runQuery with returnObject = 'result' now returns an object with 'result' and 'query'

### DIFF
--- a/box.json
+++ b/box.json
@@ -18,7 +18,7 @@
     },
     "type":"modules",
     "dependencies":{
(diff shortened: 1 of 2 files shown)


elpete:
Thanks for the PR.  This change will be part of the `2.0.0` version being worked on right now.  Look for it there!

PR #17, opened 2018-12-18
Change BaseEntity function deepEqual() to use compare() on strings

richardherbert:
When checking for equality of strings, the "==" or "eq" operator doesn't distinguish between case. As a result, "A" == "a" would return "true". 

Therefore the "==" test has been changed to compare(...) == 0 which does consider the case of the strings. As a result, compare( "A", "a") would return "-1", which means the two strings are not identical, even in case. Where as compare( "A", "A") would return "0".

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -24,7 +24,7 @@ component accessors="true" {
     property name="key"             default="id";
     property name="attributes";
     property name="meta";
-    
+
     /*=====================================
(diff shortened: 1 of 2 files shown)


elpete:
Can you submit this against `development`?  Thanks

richardherbert:
Okay, have done but I branched from your `master` so there could be some issues. You can just cherry pick may changes to merge in.

PR #22, opened 2019-03-10
Resolves QUICK-37 - add Relationship get() to wrap getResults()

MordantWastrel:
Resolves https://ortussolutions.atlassian.net/projects/QUICK/issues/QUICK-37


MordantWastrel:
I didn't add them to BaseEntity because not every relationship implements `getResults()` (even if nearly all of them do). If it's OK to add it there anyway, I'll do that and item 2 is no problem.

MordantWastrel:
The tests pass for me in ACF11 and ACF2016, so I'm not sure what's going on with Travis.

PR #25, opened 2019-03-20
update to work with `update` and `insert` properties

blusol850:
and possibly pass tests??


elpete:
I need some tests for these, @blusol850. 😄

blusol850:
@elpete tests have been added.

elpete:
Thanks, @blusol850!

PR #30, opened 2019-05-13
Add test showing subSelect global scope constraint issue, correct README docker instructions

MordantWastrel:


elpete:
Thanks for the repro PR.  This will help us decide how we want to move forward.

This issue at hand is a subselect property isn't available to `where` or `orderBy` or `groupBy`.  This is normal SQL.  To constrain a subselect property, you would need to repeat the subquery in the `where` clause.

```
SELECT `users`.*,
    ( SELECT `name`
      FROM `countries`
      WHERE `users`.`country_id` = `countries`.`id`
      LIMIT 1 ) AS `countryName`
FROM `users`
WHERE ( SELECT `name`
      FROM `countries`
      WHERE `users`.`country_id` = `countries`.`id`
      LIMIT 1 ) IS NOT NULL
```

Quick could potentially help make that easier by automatically repeating the subselect where needed.  Before that happens qb would [need better subquery support in some of its `where` methods.](https://github.com/coldbox-modules/qb/issues/67)

Another approach is to use Common Table Expressions (CTE).  These create a temporary table to query against.  It could look something like the following:

```
WITH cte_users AS (
    SELECT `users`.*,
        ( SELECT `name`
          FROM `countries`
          WHERE `users`.`country_id` = `countries`.`id`
          LIMIT 1 ) AS `countryName`
    FROM `users`
)
SELECT *
FROM cte_users
WHERE `countryName` IS NOT NULL
```

This solves the issue of repeating the subquery definition but at the cost of introducing a wrapping table.  Additionally, it would need to detect which constraints would/could be applied to the original table and which to the CTE.

Both solutions seem to have pros and cons.  For now, the workaround is to constrain the query manually again.  Creating a helper function that returns the subquery could help.

MordantWastrel:
Do CTEs create issues with grammar-specific support? I know they're easy in MSSQL but I don't know about the other grammars. 

I think I'd just go with whichever performed better. When we do this 'manually' we just do a join, which probably isn't any faster than either of the two scenarios above.

elpete:
We have this over in the issues.  We'll decide how we want to handle it over there.

PR #33, opened 2019-05-31
Fix Getting Started 404

pfreitag:
Because there is no master branch in the quick-docs repo it was causing a 404, so linking to https://quick.ortusbooks.com/getting-started instead, as that seams to be the place you should go.


elpete:
Thanks!

PR #35, opened 2019-07-03
Ignore Non-Existent Attributes

richardherbert:
Passing the complete `rc` scope to the methods `fill()`, `create()` and `update()` fails due to there being elements in the structure, such as event, that do not exist as properties of the entity. This feature enhancement adds a new argument, `ignoreNonExistentAttributes`, which defaults to `false` and therefore is a non-breaking change. When set to `true`, these methods will ignore any attributes that do not match to a property of the entity.

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -144,15 +144,18 @@ component accessors="true" {
         return this;
     }
 
-    function fill( attributes ) {
+    function fill( attributes, ignoreNonExistentAttributes = false ) {
         for ( var key in arguments.attributes ) {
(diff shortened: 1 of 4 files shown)


elpete:
Thanks, @richardherbert.  Could you please update the docs as well? https://github.com/ortus-docs/quick-docs/blob/2.2.0/getting-started/creating-new-entities.md

richardherbert:
For sure! 

I wasn't sure what the workflow was for documenting new features.

richardherbert:
Pull request on the docs raised.

PR #109, opened 2020-08-05
Feature/enable child subclass entities

jclausen:
This pull request implements the ability to apply an object-oriented approach to database design by allowing for discriminated and child class entities as Quick entities.  This feature is implemented in CFORM but not in Quick.

A separate documentation pull request will be forthcoming.  

In short:

By adding a `joincolumn` component attribute to an entity which extends another entity, the child class' table will extend the parent's table.  When loading the child class all attributes from the parent will also be loaded.

By adding a `discriminatorColumn` to the parent and a `discriminatorValue` to the child class, the entity will become discriminated with the discriminator column auto-populated on the parent.  When parents are loaded and a discriminated child exists, the child entity will be provided in collections and individual entity requests.

### DIFF
--- a/ModuleConfig.cfc
+++ b/ModuleConfig.cfc
@@ -1,86 +1,88 @@
 component {
 
-    this.name = "quick";
-    this.author = "Eric Peterson";
-    this.webUrl = "https://github.com/coldbox-modules/quick";
-    this.dependencies = [ "qb", "str", "mementifier" ];
-    this.cfmapping = "quick";
+	this.name         = "quick";
+	this.author       = "Eric Peterson";
+	this.webUrl       = "https://github.com/coldbox-modules/quick";
+	this.dependencies = [ "qb", "str", "mementifier" ];
+	this.cfmapping    = "quick";
 
-    function configure() {
-        settings = {
-            "defaultGrammar" = "AutoDiscover@qb",
-            "preventDuplicateJoins" = true,
-            "metadataCache" = {
-                "name"                  : "quickMeta",
-                "provider"              : "coldbox.system.cache.providers.CacheBoxColdBoxProvider",
-                "properties"            : {
-                    "objectDefaultTimeout"  : 0, // no timeout
-                    "useLastAccessTimeouts" : false, // no last access timeout
-                    "maxObjects"            : 300,
-                    "objectStore"           : "ConcurrentStore"
-                }
+	function configure() {
+		settings = {
+			"defaultGrammar"        : "AutoDiscover@qb",
+			"preventDuplicateJoins" : true,
+			"metadataCache"         : {
+				"name"       : "quickMeta",
+				"provider"   : "coldbox.system.cache.providers.CacheBoxColdBoxProvider",
+				"properties" : {
+					"objectDefaultTimeout"  : 0, // no timeout
+					"useLastAccessTimeouts" : false, // no last access timeout
+					"maxObjects"            : 300,
+					"objectStore"           : "ConcurrentStore"
+				}
 			}
-        };
+		};
 
-        interceptorSettings = {
-            customInterceptionPoints = [
-                "quickInstanceReady",
-                "quickPreLoad",
-                "quickPostLoad",
-                "quickPreSave",
-                "quickPostSave",
-                "quickPreInsert",
-                "quickPostInsert",
-                "quickPreUpdate",
-                "quickPostUpdate",
-                "quickPreDelete",
-                "quickPostDelete"
-            ]
-        };
+		interceptorSettings = {
+			customInterceptionPoints : [
+				"quickInstanceReady",
+				"quickPreLoad",
+				"quickPostLoad",
+				"quickPreSave",
+				"quickPostSave",
+				"quickPreInsert",
+				"quickPostInsert",
+				"quickPreUpdate",
+				"quickPostUpdate",
+				"quickPreDelete",
+				"quickPostDelete"
+			]
+		};
 
-        binder.map( "quick.models.BaseEntity" )
-            .to( "#moduleMapping#.models.BaseEntity" );
+		binder.map( "quick.models.BaseEntity" ).to( "#moduleMapping#.models.BaseEntity" );
 
-        binder.getInjector().registerDSL( "quickService", "#moduleMapping#.dsl.QuickServiceDSL" );
-    }
+		binder.getInjector().registerDSL( "quickService", "#moduleMapping#.dsl.QuickServiceDSL" );
+	}
 
-    function onLoad() {
-        binder.map( alias = "QuickBuilder@quick", force = true )
-            .to( "#moduleMapping#.models.QuickBuilder" )
-            .initArg( name = "grammar", dsl = settings.defaultGrammar )
-            .initArg( name = "preventDuplicateJoins", value = settings.preventDuplicateJoins )
-            .initArg( name = "utils", dsl = "QueryUtils@qb" )
-            .initArg( name = "returnFormat", value = "array" );
+	function onLoad() {
+		binder
+			.map( alias = "QuickBuilder@quick", force = true )
+			.to( "#moduleMapping#.models.QuickBuilder" )
+			.initArg( name = "grammar", dsl = settings.defaultGrammar )
+			.initArg( name = "preventDuplicateJoins", value = settings.preventDuplicateJoins )
+			.initArg( name = "utils", dsl = "QueryUtils@qb" )
+			.initArg( name = "returnFormat", value = "array" );
 
-        if( isSimpleValue( settings.metadataCache ) ){
-            settings.metadataCache = { "name" : settings.metadataCache };
-        }
+		if ( isSimpleValue( settings.metadataCache ) ) {
+			settings.metadataCache = { "name" : settings.metadataCache };
+		}
 
-        if( settings.metadataCache.name != "quickMeta" ){
-            binder.map( "cachebox:quickMeta" )
-                    .toDSL( "cachebox:#settings.metadataCache.name#" );
-        } else {
-            controller.getCachebox().createCache( argumentCollection = settings.metadataCache );
-        }
-        
-    }
+		if ( settings.metadataCache.name != "quickMeta" ) {
+			binder.map( "cachebox:quickMeta" ).toDSL( "cachebox:#settings.metadataCache.name#" );
+		} else {
+			controller.getCachebox().createCache( argumentCollection = settings.metadataCache );
+		}
+	}
 
-    function onUnload(){
-        controller.getCachebox().getCache( settings.metadataCache.name ).clearAll();
-    }
+	function onUnload() {
+		controller
+			.getCachebox()
+			.getCache( settings.metadataCache.name )
+			.clearAll();
+	}
 
 
-    /**
-     * This interceptor ensures that the `_mapping` property for a Quick entity
-     * is correctly set to the mapping used in WireBox.
-     *
-     * @event          The current request context
-     * @interceptData  The intercept data for `afterInstanceAutowire`.
-     *                 { mapping, target, targetID, injector }
-     */
-    function beforeInstanceAutowire( event, interceptData ) {
-        if ( structKeyExists( interceptData.target, "isQuickEntity" ) ) {
-            interceptData.target.set_mapping( interceptData.targetID );
-        }
-    }
+	/**
+	 * This interceptor ensures that the `_mapping` property for a Quick entity
+	 * is correctly set to the mapping used in WireBox.
+	 *
+	 * @event          The current request context
+	 * @interceptData  The intercept data for `afterInstanceAutowire`.
+	 *                 { mapping, target, targetID, injector }
+	 */
+	function beforeInstanceAutowire( event, interceptData ) {
+		if ( structKeyExists( interceptData.target, "isQuickEntity" ) ) {
+			interceptData.target.set_mapping( interceptData.targetID );
+		}
+	}
+
 }
--- a/ModuleConfig.cfc
+++ b/ModuleConfig.cfc
@@ -43,6 +43,42 @@ component {
             .initArg( name = "returnFormat", value = "array" );
     }
 
+    function onUnload(){
+        structDelete( application, "quickMeta" );
+    }
+
+    /**
+     * This interception ensures that quick entity parent/child relationships are available to discriminated entities
+     * @event          The current request context
+     * @interceptData  The intercept data for `afterAspectsLoad`.
+     */
+    function afterAspectsLoad( event, interceptData ){
+        param application.quickMeta = {};
+        param application.quickMeta.discriminators = {};
+        
+        binder.getMappings().each( function( key, mapping ){
+            if( !mapping.isDiscovered() ){
+                mapping.process( binder, application.wirebox );
+            }
+            var meta = mapping.getObjectMetadata();
+            if( structKeyExists( meta, "joincolumn" ) && structKeyExists( meta, "discriminatorValue" ) ){
+                // retrieve the inheritance metadata
+                meta = getComponentMetadata( meta.fullName );
+                parentMeta = meta.extends;
+                
+                if( !structKeyExists( parentMeta, "table" ) ) parentMeta = application.wirebox.getInstance( parentMeta.fullName ).get_Meta();
+                if( !structKeyExists( application.quickMeta.discriminators, parentMeta.table ) ){
+                    application.quickMeta.discriminators[ parentMeta.table ] = {};
+                }
+                application.quickMeta.discriminators[ parentMeta.table ][ meta.discriminatorValue ] = {
+                    "mapping" : meta.fullName,
+                    "joincolumn" : meta.joinColumn
+                };
+            }
+        } );
+        
+    }
+
     /**
      * This interceptor ensures that the `_mapping` property for a Quick entity
      * is correctly set to the mapping used in WireBox.
--- a/ModuleConfig.cfc
+++ b/ModuleConfig.cfc
@@ -62,17 +62,28 @@ component {
             }
             var meta = mapping.getObjectMetadata();
             if( structKeyExists( meta, "joincolumn" ) && structKeyExists( meta, "discriminatorValue" ) ){
-                // retrieve the inheritance metadata
-                meta = getComponentMetadata( meta.fullName );
-                parentMeta = meta.extends;
+                // retrieve the inheritance metadata and attributes
+                var childClass = application.wirebox.getInstance( meta.fullName );
+                var childAttributes = childClass.get_Attributes().reduce( function( acc, attr, data ){ 
+                    if( !data.isParentColumn && !data.virtual && !data.exclude ){
+                        acc.append( data );
+                    }
+                    return acc;
+                 }, [] );
+                var meta = childClass.get_Meta();
+                var parentMeta = meta.parentEntity.meta;
                 
                 if( !structKeyExists( parentMeta, "table" ) ) parentMeta = application.wirebox.getInstance( parentMeta.fullName ).get_Meta();
+                
                 if( !structKeyExists( application.quickMeta.discriminators, parentMeta.table ) ){
                     application.quickMeta.discriminators[ parentMeta.table ] = {};
                 }
-                application.quickMeta.discriminators[ parentMeta.table ][ meta.discriminatorValue ] = {
+
+                application.quickMeta.discriminators[ parentMeta.table ][ meta.parentEntity.discriminatorValue ] = {
                     "mapping" : meta.fullName,
-                    "joincolumn" : meta.joinColumn
+                    "table" : meta.table,
+                    "joincolumn" : meta.parentEntity.joinColumn,
+                    "attributes" : childAttributes
                 };
             }
         } );
--- a/ModuleConfig.cfc
+++ b/ModuleConfig.cfc
@@ -71,7 +71,7 @@ component {
                     return acc;
                  }, [] );
                 var meta = childClass.get_Meta();
-                var parentMeta = meta.parentEntity.meta;
+                var parentMeta = meta.parentDefinition.meta;
                 
                 if( !structKeyExists( parentMeta, "table" ) ) parentMeta = application.wirebox.getInstance( parentMeta.fullName ).get_Meta();
                 
@@ -79,10 +79,10 @@ component {
                     application.quickMeta.discriminators[ parentMeta.table ] = {};
                 }
 
-                application.quickMeta.discriminators[ parentMeta.table ][ meta.parentEntity.discriminatorValue ] = {
+                application.quickMeta.discriminators[ parentMeta.table ][ meta.parentDefinition.discriminatorValue ] = {
                     "mapping" : meta.fullName,
                     "table" : meta.table,
-                    "joincolumn" : meta.parentEntity.joinColumn,
+                    "joincolumn" : meta.parentDefinition.joinColumn,
                     "attributes" : childAttributes
                 };
             }
--- a/ModuleConfig.cfc
+++ b/ModuleConfig.cfc
@@ -9,7 +9,17 @@ component {
     function configure() {
         settings = {
             "defaultGrammar" = "AutoDiscover@qb",
-            "preventDuplicateJoins" = true
+            "preventDuplicateJoins" = true,
+            "metadataCache" = {
+                "name"                  : "quickMeta",
+                "provider"              : "coldbox.system.cache.providers.CacheBoxColdBoxProvider",
+                "properties"            : {
+                    "objectDefaultTimeout"  : 0, // no timeout
+                    "useLastAccessTimeouts" : false, // no last access timeout
+                    "maxObjects"            : 300,
+                    "objectStore"           : "ConcurrentStore"
+                }
+			}
         };
 
         interceptorSettings = {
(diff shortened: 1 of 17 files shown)


tests/specs/integration/BaseEntity/Relationships/EagerLoadingSpec.cfc
```diff
@@ -274,6 +274,10 @@ component extends="tests.resources.ModuleIntegrationSpec" appMapping="/app" {
 			} );
 
 			it( "can eager load polymorphic has many relationships", function() {
+				// delete our internal comments to allow the test to pass:
+				getInstance( "InternalComment" ).get().each( function( comment ){ comment.delete(); } );
```
elpete:
```suggestion
				getInstance( "InternalComment" ).deleteAll();
```

tests/specs/integration/BaseEntity/Relationships/EagerLoadingSpec.cfc
```diff
@@ -295,6 +299,9 @@ component extends="tests.resources.ModuleIntegrationSpec" appMapping="/app" {
 			} );
 
 			it( "can eager load a nested relationship", function() {
+				// delete our internal comments to allow the test to pass:
+				getInstance( "InternalComment" ).get().each( function( comment ){ comment.delete(); } );
```
elpete:
```suggestion
				getInstance( "InternalComment" ).deleteAll();
```

tests/specs/integration/BaseEntity/Relationships/PolymorphicHasManySpec.cfc
```diff
@@ -3,6 +3,10 @@ component extends="tests.resources.ModuleIntegrationSpec" appMapping="/app" {
 	function run() {
 		describe( "Polymorphic Has Many Spec", function() {
 			it( "can get the related polymorphic entities", function() {
+				// delete our internal comments to allow the test to pass:
+				getInstance( "InternalComment" ).get().each( function( comment ){ comment.delete(); } );
```
elpete:
```suggestion
				getInstance( "InternalComment" ).deleteAll();
```

tests/specs/integration/BaseEntity/Relationships/RelationshipsCountSpec.cfc
```diff
@@ -12,6 +12,11 @@ component extends="tests.resources.ModuleIntegrationSpec" appMapping="/app" {
 			} );
 
 			it( "can add a subselect for the count of a relationship without loading the relationship", function() {
+
+				// delete our internal comments to allow the test to pass:
+				getInstance( "InternalComment" ).get().each( function( comment ){ comment.delete(); } );
```
elpete:
```suggestion
				getInstance( "InternalComment" ).deleteAll();
```

tests/specs/integration/BaseEntity/Relationships/RelationshipsCountSpec.cfc
```diff
@@ -59,7 +64,10 @@ component extends="tests.resources.ModuleIntegrationSpec" appMapping="/app" {
 				expect( posts[ 4 ].getPost_Pk() ).toBe( 7777 );
 			} );
 
-			it( "can add multiple counts at once", function() {
+			it( "can add multiple counts at once", function() {// delete our internal comments to allow the test to pass:
+				getInstance( "InternalComment" ).get().each( function( comment ){ comment.delete(); } );
```
elpete:
```suggestion
				getInstance( "InternalComment" ).deleteAll();
```

tests/specs/integration/BaseEntity/Relationships/RelationshipsCountSpec.cfc
```diff
@@ -110,6 +118,10 @@ component extends="tests.resources.ModuleIntegrationSpec" appMapping="/app" {
 			} );
 
 			it( "can constrain counts at runtime", function() {
+				// delete our internal comments to allow the test to pass:
+				getInstance( "InternalComment" ).get().each( function( comment ){ comment.delete(); } );
```
elpete:
```suggestion
				getInstance( "InternalComment" ).deleteAll();
```

tests/specs/integration/BaseEntity/Relationships/RelationshipsCountSpec.cfc
```diff
@@ -148,6 +160,10 @@ component extends="tests.resources.ModuleIntegrationSpec" appMapping="/app" {
 			} );
 
 			it( "can alias the counts attribute name", function() {
+				// delete our internal comments to allow the test to pass:
+				getInstance( "InternalComment" ).get().each( function( comment ){ comment.delete(); } );
```
elpete:
```suggestion
				getInstance( "InternalComment" ).deleteAll();
```

models/BaseEntity.cfc
```diff
@@ -444,7 +452,13 @@ component accessors="true" {
 			if ( value.virtual && !withVirtualAttributes ) {
 				return items;
 			}
-			items.append( asColumnNames ? value.column : key );
+			items.append( 
+				asColumnNames 
+				? value.isParentColumn 
+					? ( variables._parentEntity.meta.table & "." & value.column )
```
elpete: Is `_parentEntity` an actual entity instance?  If so, then:
```suggestion
					? variables._parentEntity.qualifyColumn( value.column ) 
```

models/BaseEntity.cfc
```diff
@@ -1041,8 +1057,10 @@ component accessors="true" {
 			}
 		);
 		activateGlobalScopes();
-		var data = retrieveQuery()
-			.from( tableName() )
+
+		var q = variables._hasParentEntity ? newQuery() : retrieveQuery();
```
elpete, line 1061: I'm not sure this is correct.  Calling `newQuery` here will remove any constraints already added to the query, e.g. `user.where( "active", 1 ).orderBy( "createdDate" ).find( 1 )` will lose the `where` and `orderBy`.

models/BaseEntity.cfc
```diff
@@ -1165,10 +1183,36 @@ component accessors="true" {
...
+				data[ variables._meta.localMetadata.discriminatorColumn ]
+			)
+		) {
+			var childClass = variables._wirebox.getInstance(
+				application.quickMeta.discriminators[ variables._meta.table ][
+					data[ variables._meta.localMetadata.discriminatorColumn ]
+				].mapping
+			);
+
+			keyNames().each( function( key, i ) {
```
elpete, line 1205: I've been told lately to avoid `arrayEach` and `structEach` for memory issues in Lucee.  Know anything about that?

elpete, line 1205: Can you help me understand what this section is doing?  I'm feeling a bit confused.  It feels like it would override ids, so I must be missing something.

models/BaseEntity.cfc
```diff
@@ -1165,10 +1183,36 @@ component accessors="true" {
...
+				data[ childClass.keyNames()[ i ] ] = data[ key ];
+			} );
+
+			return childClass.hydrate( data, true );
```
elpete, line 1209: One thing missing when using `hydrate` is `assignOriginalAttributes`.  Let's add that in this PR and both paths can use `hydrate`.

models/BaseEntity.cfc
```diff
@@ -1322,6 +1366,23 @@ component accessors="true" {
...
+			if ( isLoaded() ) {
+				var parent = variables._wirebox
+					.getInstance( variables._parentEntity.meta.fullName )
+					.set_LoadChildren( false )
+					.findOrFail( keyValues() );
+			} else {
+				var parent = variables._wirebox.getInstance( variables._parentEntity.meta.fullName );
+			}
+
+			parent.fill( getMemento(), true ).save();
```
elpete, line 1379: I wonder if we could use [`updateOrCreate`](https://quick.ortusbooks.com/guide-1/getting-started/creating-new-entities#updateorcreate) here?

models/BaseEntity.cfc
```diff
@@ -1395,6 +1461,15 @@ component accessors="true" {
 				} );
 			} )
 			.delete();
+
+		if ( variables._hasParentEntity ) {
+			variables._wirebox
+				.getInstance( variables._parentEntity.meta.fullName )
+				.set_LoadChildren( false )
+				.findOrFail( keyValues() )
```
elpete, line 1469: It would be nice if we could skip loading the entity here and just delete it.

models/BaseEntity.cfc
```diff
@@ -2380,14 +2455,51 @@ component accessors="true" {
...
-			.select( retrieveQualifiedColumns() );
+			.from( tableName() );
+
+		if ( variables._hasParentEntity ) {
+			return variables._builder
+				.newQuery()
+				.select( retrieveQualifiedColumns() )
+				.join(
+					variables._parentEntity.meta.table,
+					variables._parentEntity.meta.table & "." & variables._parentEntity.key,
```
elpete, line 2471: Again, I would love to use the parent entity's `qualifyColumn` method here somehow.  Not essential, but it would be nice.

models/BaseEntity.cfc
```diff
@@ -2380,14 +2455,51 @@ component accessors="true" {
...
+					this.qualifyColumn( variables._key )
+				);
+		} else if (
+			variables._meta.keyExists( "discriminatorColumn" ) && application.quickMeta.discriminators.keyExists(
+				tableName()
+			)
+		) {
+			var q = variables._builder..newQuery();
```
elpete, line 2479: Extra `.` here.

```suggestion
			var q = variables._builder.newQuery();
```

models/BaseEntity.cfc
```diff
@@ -2380,14 +2455,51 @@ component accessors="true" {
...
+			return variables._builder
+				.newQuery()
+				.select( retrieveQualifiedColumns() )
+				.join(
+					variables._parentEntity.meta.table,
+					variables._parentEntity.meta.table & "." & variables._parentEntity.key,
+					this.qualifyColumn( variables._key )
+				);
+		} else if (
+			variables._meta.keyExists( "discriminatorColumn" ) && application.quickMeta.discriminators.keyExists(
```
elpete, line 2475: The way I understand this is

If this entity has a discriminator column
and if we know of a parent entity for this entity's table,
then do stuff....

But what if we have a `discriminatorColumn` and we _don't_ know of a parent entity for this entity's table?  Shouldn't that be an exception with a helpful error message?

elpete: In fact `joincolumn`, `discriminatorColumn`, `discriminatorValue`, and `extends` are all required if one is provided.  I think we can throw if that's not the case when inspecting the metadata with a helpful explanation on how to fix it.

models/BaseEntity.cfc
```diff
@@ -3008,23 +3148,45 @@ component accessors="true" {
...
+		variables._fullName        = variables._meta.fullName;
+		variables._entityName      = variables._meta.entityName;
+		variables._table           = variables._meta.table;
+		variables._hasParentEntity = variables._meta.hasParentEntity;
+
+		if ( variables._hasParentEntity ) {
+			variables._parentEntity = variables._meta.parentEntity;
+
+			if ( structKeyExists( variables._parentEntity, "discriminatorColumn" ) ) {
+				variables[ "get" & variables._parentEntity.discriminatorColumn ] = function() {
```
elpete, line 3180: Instead of adding it here as a function, let's add a check to `retrieveAttribute` or in our `tryAttributesGetter` call.  This way a user can still override the function if they would like.  If we need to call the getter ourselves internally, we should probably store this value in some private state (prefixed with an `_`).

models/BaseEntity.cfc
```diff
@@ -3421,6 +3618,8 @@ component accessors="true" {
 		any value,
 		boolean checkNullValues = true
 	) {
+		arguments.column = arguments.column;
```
elpete, line 3621: Don't think this line is needed.

models/BaseEntity.cfc
```diff
@@ -3583,8 +3783,9 @@ component accessors="true" {
 	 */
 	private boolean function canInsertAttribute( required string name ) {
 		var alias = retrieveAliasForColumn( arguments.name );
-		return variables._meta.attributes.keyExists( alias ) &&
-		variables._meta.attributes[ alias ].insert;
+		return variables._attributes.keyExists( alias ) &&
+		variables._attributes[ alias ].insert &&
+		!variables._attributes[ alias ].isParentColumn;
```
elpete, line 3788: This is fine as well, but any reason we don't pass `insert: false` and `update: false` when creating a parent column?

elpete:
> It's soooo much prettier with indents... ;)

Sounds more like a cfformat bug than desired formatting.

models/QuickBuilder.cfc
```diff
@@ -673,7 +673,9 @@ component extends="qb.models.Query.QueryBuilder" accessors="true" {
 			return arguments.column;
 		}
 
-		return listLast( getFrom(), " " ) & "." & getEntity().retrieveColumnForAlias( arguments.column );
+		return getEntity().isParentAttribute( arguments.column )
+				? getEntity().retrieveColumnForAlias( arguments.column )
```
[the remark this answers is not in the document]
jclausen: I had that in there before and can't remember why I took it out now. It might not be a bad idea if the child class has a column that overloads the parent.  I'm not sure how qB data will normalize that column name in the query, though.  I'll make the change after a few tests.

ModuleConfig.cfc
```diff
@@ -43,6 +43,53 @@ component {
...
+    function onUnload(){
+        structDelete( application, "quickMeta" );
+    }
+
+    /**
+     * This interception ensures that quick entity parent/child relationships are available to discriminated entities
+     * @event          The current request context
+     * @interceptData  The intercept data for `afterAspectsLoad`.
+     */
+    function afterAspectsLoad( event, interceptData ){
```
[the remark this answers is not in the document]
jclausen: The problem with the using the metadata cache is that it expires - because it uses the `default` cache.  I thought about creating a custom quick cachebox cache that was scoped in to the application.  Ultimately, though, I don't want anything in this metadata to ever be evicted, because I need the parent entities to be able to access it at all times.   Storing it in the app scope guarantees it's always there.

tests/resources/app/models/Comment.cfc
```diff
@@ -1,4 +1,8 @@
-component extends="quick.models.BaseEntity" accessors="true" {
+component 
+    extends="quick.models.BaseEntity"
+    discriminatorColumn="designation"
```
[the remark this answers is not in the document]
jclausen: It's appended to the attributes in the metadata inspection of the `BaseEntity`, along with a getter.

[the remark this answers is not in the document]
jclausen: Having it on the parent is currently the way I detect whether to look for and load child entities and, also, to make it an a data attribute of the parent.  

It also matches the pattern that CF ORM uses and, if you had to declare it on the child, would be duplication of code throughout any subclass entities.

tests/specs/integration/BaseEntity/ChildClassSpec.cfc
```diff
@@ -0,0 +1,297 @@
...
+                        "title" : "Give me a Break",
+                        "downloadURL" : "https://www.youtube.com/watch?v=0nkcVz1mad0",
+                        "createdDate" : now(),
+                        "modifiedDate" : now(),
+                        "catchiness" : 2
+                    }
+                ).save();
+
+                var memento = newJingle.getMemento();
+
```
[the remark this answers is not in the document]
jclausen: Can do

models/BaseEntity.cfc
jclausen: A fix for this typo has been pushed.

models/BaseEntity.cfc
jclausen: `discriminatorColumn` is always on the parent ( see above ).  `discriminatorValue` is always on the child, as is `extends` and `joinColumn`.  

Since the application discriminator data is detected when inspecting the _children_, if the key is not in the application scope, there are no registered children and we would fall through to to the normal behavior of an entity without parents or children.

models/BaseEntity.cfc
jclausen: Good call.  I'll take a look at how that's structured and move it over.

models/BaseEntity.cfc
jclausen: I do believe you are correct! :)

models/BaseEntity.cfc
jclausen: Because we actually do insert _and_ update.  See the `save()` method updates

jclausen:
> This is going to be so awesome! I have made notes of some changes and questions.
> 
> One overall question / idea I have to make this more Quick-like is to perhaps use a function to do the `discriminatorColumn`, `joinColumn`, and `discriminatorValue` constraints. This would allow us to do any sort of constraints for the join, not just a simple column-to-value constraint. I'm not sure the best idea for the function name or names, but I like the flexibility of the approach and that it feels more in line with `relationships`. I'm interested in your thoughts on the idea.
> 
> I am very happy to work on these features with you as well. Let's chat about the best way to do that.

jclausen:
Ugh.  Sorry, I accidentally hit "close" on a comment.

jclausen:
Let's discuss the function idea a bit.  I'm not a fan of using functions to designate these values, because doing so requires more metadata inspection on data which would "just be there" with a shallow metadata peek done by Wirebox.  Given that the relationships in child entities and discriminators are true inheritance, I feel like it makes sense and is good coding practice to declare these attributes on the component.

models/QuickBuilder.cfc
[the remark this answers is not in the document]
jclausen: Done

tests/specs/integration/BaseEntity/ChildClassSpec.cfc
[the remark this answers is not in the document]
jclausen: Persistence equality test added.

tests/specs/integration/BaseEntity/MetadataSpec.cfc
```diff
@@ -64,6 +64,45 @@ component extends="tests.resources.ModuleIntegrationSpec" appMapping="/app" {
...
+												.toBeStruct()
+												.toHaveKey( "meta" )
+												.toHaveKey( "joincolumn" )
+												.toHaveKey( "discriminatorValue" )
+												.toHaveKey( "discriminatorColumn" );
+
+				} );
+
+				it( "can detect a non-discriminated child subclass", function(){
+					var discriminated = getInstance( "Jingle" );
```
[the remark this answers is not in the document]
jclausen: `Jingle` is a non-discriminated child class.  Note the `notToHaveKey` below

tests/specs/integration/BaseEntity/ChildClassSpec.cfc
```diff
@@ -0,0 +1,297 @@
...
+                    {
+                        "title" : "Give me a Break",
+                        "downloadURL" : "https://www.youtube.com/watch?v=0nkcVz1mad0",
+                        "createdDate" : now(),
+                        "modifiedDate" : now(),
+                        "catchiness" : 2
+                    }
+                ).save();
+
+                transactionSetSavePoint( 'jingle-saved' );
```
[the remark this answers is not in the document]
jclausen: It's probably not necessary with quick since it executes against the database immediately and doesn't pool statements, actually. With CF ORM, you need them to make sure the SQL is executed within the transaction so that the database is up to date.  I can remove them from these tests.

tests/specs/integration/BaseEntity/ChildClassSpec.cfc
```diff
@@ -0,0 +1,297 @@
...
+                        "commentableId" : 1345,
+                        "commentableType" : 'Post',
+                        "userId" : getInstance( "User" ).first().getId(),
+                        "createdDate" : now(),
+                        "modifiedDate" : now()
+                    }
+                ).save();
+
+                var parentRetrieval = getInstance( "Comment" ).find( newComment.getId() );
+                expect( isInstanceOf( parentRetrieval, "InternalComment" ) ).toBeTrue();
```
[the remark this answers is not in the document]
jclausen: I'll have to see if that works.  I've had issues with `toBeInstanceOf` requiring a full mapping to return positive.

tests/specs/integration/BaseEntity/ChildClassSpec.cfc
```diff
@@ -0,0 +1,297 @@
...
+            });
+
+            it( "Will return an array of child classes when fetching multiple results from the parent class", function(){
+                var internalComments = getInstance( "Comment" ).where( 'designation', 'internal' ).get();
+                internalComments.each( function( comment ){
```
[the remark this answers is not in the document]
jclausen: Yes, it will give you a mix of classes, if you just did an `all()`.  In this case, you would get some `Comment` entities and 1 `InternalComment` entity.

models/BaseEntity.cfc
jclausen: No `_parentEntity` is the parent's definition with a nested `_meta` key.

models/BaseEntity.cfc
jclausen: I probably need to double back on this, then, as `newQuery` was where I could place all of the join logic for parents and children without having to duplicate it all over.  `retreiveQuery`  can't be used because it affects the subqueries.  Currently, I don't use `newQuery` for parent/child entities, then the joins for the parent attributes don't happen.

models/BaseEntity.cfc
jclausen: No, I've heard about massive memory leaks with closures which were `var` scoped so I suppose that might apply to the closure used inside of `each`.  Sounds like something Lucee needs to fix, though.  I've found member function loops to, generally, be faster.

jclausen: The key for the child class is the join column, so this is taking the parent keys and assigning the join column names in to the data structure used to hydrate the child.  Without this code, the call to `hydrate` throws and exception because the keys aren't present.

models/BaseEntity.cfc
jclausen: Done.

models/BaseEntity.cfc
jclausen: Let me give that a try.

models/BaseEntity.cfc
jclausen: I thought about that, but my concern is that an overload to `delete` within the parent entity would be bypassed, then. If the parent entity needs to do any additional work during deletion, then this catches that.

models/BaseEntity.cfc
jclausen: See note above.  `variables._parentEntity` is a struct containing the join/discriminator info and the parent meta

models/BaseEntity.cfc
jclausen: Actually, this closure creation is not needed as the discriminator column is already scoped as a attribute and the `onMissingMethod` fall through handles it just fine. :)   Removed.

models/BaseEntity.cfc
jclausen: I just tried this and it's a no-go because `updateOrCreate` treats an empty key value ( from the getMemento() ) as the authoritative value when it fills and then the entity returns an empty string for the key.  I also can't  use it without some parent introspection beforehand and conditional struct assembly.  

I think both `updateOrCreate` and `firstOrNew` could be modified to deal with the issue, but my concern would be the possibility of bugs when dealing with different key types - especially string key types like UUID.

tests/resources/app/models/Comment.cfc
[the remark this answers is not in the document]
elpete: I'm not talking about in addition to but rather in place of.  I think it's worth a look still.

elpete:
Going through this I have three thoughts I feel like we need to resolve before merging:
1) I'd like a way to express complex relationships between parent and child.  This doesn't have to be the default.  It can be an optional method that is called that is usually empty.  The default can be the metadata.

2) A few tests around child entities calling their own custom relationships and parent relationships are probably needed.  I think we might need to make a decision about what are the default key names when using a child entity.

3) I really don't like caching the metadata upfront in the application scope.  I see the current need - getting a map of all tables and their children - but something Quick has tried to always do is be lazily evaluated.  That's one of the reasons the metadata is not processed until the entity is first created.  I agree that setting a timeout of `0` on these metadata cache calls is important, so let's add that.  I don't know what is the best way set up this join dictionary without metadata inspection.  One idea is to create a `_inheritanceMap` that maps discriminator values to mappings on the Parent entity.  Then the Parent entity can load the different types of children without having to pre-load all entities.  That does duplicate the `discriminatorValue` on the Child entity, though, so it is not without tradeoffs.  Unless there is a _very_ compelling reason, though, I'd like to avoid the processing of all metadata up front.

Let's chat about these in more detail and make a plan to fix them.  Like I mentioned, I am more than happy to help or make all the changes we agree on.

tests/resources/ModuleIntegrationSpec.cfc
```diff
@@ -5,6 +5,9 @@ component extends="coldbox.system.testing.BaseTestCase" {
 
         getController().getModuleService()
             .registerAndActivateModule( "quick", "testingModuleRoot" );
+        
+        getController().getInterceptorService()
```
elpete, line 9: I think we can remove this with the definition.

models/BaseEntity.cfc
```diff
@@ -3140,9 +3267,97 @@ component accessors="true" {
...
+		return variables._meta.localMetadata.keyExists( "discriminatorColumn" )
+		&& variables._discriminators.len();
+	}
+
+	public function getParentDefinition() {
+		return hasParentEntity() ? variables._parentDefinition : javacast( "null", 0 );
+	}
+
+	public function getDiscriminations() {
+		return variables._cache.getOrSet( "quick-metadata:discriminations-#tableName()#", function() {
```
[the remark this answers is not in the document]
jclausen: Can you clarify what you mean by `mapping`?    I thought about `entityName()`, but this seemed more likely to prevent a recreation of the cache.  It's probably 6 one/ 1/2 dozen the other, though.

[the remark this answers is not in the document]
elpete: `mapping` is the mapping used to create the instance from WireBox.  It is injected in to the metadata.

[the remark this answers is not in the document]
jclausen: OK. Let me take a look at it.  My only concern would be running in to key name lengths in a non-memory cache provider. It seemed safer to use the table name. It should be able to work 9/10 times.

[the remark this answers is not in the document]
elpete: Well, maybe we run with it and come back later if it is a problem.

Issue #133, opened 2020-11-11
Eager loading throws error when using CBORMCompatEntity


jclausen:
Due to the overload of the `get` method in `CBORMCompatEntity`, the pass through on the `with( 'myRelationship'  )` when using an in-entity will always fail with the exception: `The value returned from the getEager function is not of type array.`

CBORM's behavior deviates from the overloaded `get` in `CBORMCompatEntity` in two ways:

1.  It allows the method to be run without providing an explicit identifier
2.  It returns a blank entity instead of a null.

As a result of these two behaviors, any invocation of `get()` on a sub-entity will result in a single empty instance of the entity being returned.

Suggestion: allow `get()` calls without arguments to pass through to the super scope and set the default of `returnNew` to `false` to match CBORM's implementation

elpete:
Addressed by PR #322, now merged into the next branch.

PR #135, opened 2020-11-11
WIP: Optimizations and Enhancements to ORM Compat and Collection Loading

jclausen:
- Resolves Issue #133 - Allows the `get` method to pass through to the super scope when the identifier argument is null.  Sets the `returnNew` default to `false` to match the behavior of `cborm`
- Makes the Cast objects singletons to speed up load times
- Stores a reference prototype state and bypasses injection when loading a nested collection of the same objects.  In tests this reduces a collection load of 600+ items from 160s to 3.5s


jclausen:
Marking this as WIP for the moment, as I continue to work through some concerns with the scope injection

PR #178, opened 2022-04-29
Allow Non-Loaded Entities to Return Null Values for Relationships

homestar9:
This change should (hopefully) be pretty straightforward. In order to be able to retrieve null values back for relationships on a non-loaded (new) entity, all of the `guardAgainstNotLoaded()` checks were removed.  I also applied the same fix to `isNullValue()` as you made in Quick version 5

The rest of the `guardAgainstNotLoaded()` checks are still in place so you shouldn't be able to create new relationships on a non-loaded entity.


bdw429s:

https://community.ortussolutions.com/t/quick-4-2-4-this-instance-is-not-loaded-so-it-cannot-access-the-relationship/9216/1

PR #182, opened 2022-05-27
Add Wirebox `include` and `exclude` arguments

homestar9:
You can now include or exclude fields based similar to Wirebox's `populateFromStruct()`.

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -597,8 +597,21 @@ component accessors="true" {
 	 *
 	 * @return                       quick.models.BaseEntity
 	 */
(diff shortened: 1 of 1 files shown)


models/BaseEntity.cfc
```diff
@@ -597,8 +597,21 @@ component accessors="true" {
 	 *
 	 * @return                       quick.models.BaseEntity
 	 */
-	public any function fill( struct attributes = {}, boolean ignoreNonExistentAttributes = false ) {
+	public any function fill( 
+		struct attributes = {}, 
+		boolean ignoreNonExistentAttributes = false,
+		string include = "",
```
elpete, line 603: Can we pass an array as well as a list?

PR #185, opened 2022-06-24
Allow Single Table Inheritance (STI) With Discriminated Entities

homestar9:
This PR enhances the functionality of discriminated entities and allows for child classes to get their data from the same table as the parent without the need for a join column.  This pattern is known as Single Table Inheritance (STI) and is quite common when developing apps where you do not always have control over the database or it's level or normalization.

Additionally, I also gave Quick some additional flexibility with the discriminator column so users can now use dot notation to indicate if the discriminator value should come from a different table which might be accessible via a join.


homestar9:
@elpete please let me know if you have any questions.  Thanks to @jclausen and others on Slack for helping me put this together.  All the necessary tests should be there

homestar9:
I made one minor change to better help differentiate between polymorphic discriminations vs single table discriminations.  Since polymorphic came first in Quick, if the users want to use STI, they will have to include the `singleTableInheritance=true' metadata in their base class.  This also has a nice side effect of preserving the polymorphic guard in case someone forgets to include their join column.

bdw429s:

https://community.ortussolutions.com/t/quick-4-2-4-can-discriminated-entities-originate-from-same-table/9213/2

PR #193, opened 2022-10-01
Add include and exclude arguments to fill method

homestar9:
This enhancement to `fill()` simulates Wirebox's `populateFromStruct()` ability to  `include` or `exclude` fields when populating.  @elpete, This PR was updated to allow for a simple list or an array to be passed for either argument.


homestar9:
Not sure why the Format test failed.  All looks good on my end.

homestar9:
Wohoo! Looks like all tests passed @elpete

Issue #200, opened 2022-10-14
[5.1.3] Primary Key Returning Decimal Instead of Integer


homestar9:
Upgrading from 5.1.2 to 5.1.3 introduces a bug where the primary key returns a decimal instead of an integer when getting the memento of an entity that was just **created**. 

**5.1.2:**
![image](https://user-images.githubusercontent.com/5341142/195888265-0dd98058-c2a4-44f5-8277-f4bd8f9a1bd9.png)

**5.1.3:**
![image](https://user-images.githubusercontent.com/5341142/195888040-ea42f344-312a-4df3-b9ad-fe6db6a16f1d.png)

Engine Tested: ACF 2018.0.14 and ACF 2018.0.15
DB Engine: MSSQL Server

Interestingly, in 5.1.2, the memento returns a string instead of a number, as noted by the value wrapped in double quotes.  Perhaps the intent of 5.1.3 was to ensure the value gets returned in numeric form.  However, ACF is including the decimal for some reason which will throw off any app that expects the primary key to be an integer.

I will report back when I have more information but wanted to let @elpete know since this feels like a big one IMO.

homestar9:
Confirmed. The issue is with ACF only (thanks Adobe) and has to do with how ACF handles `val()` which was introduced in
https://github.com/coldbox-modules/quick/commit/1092a8e5c633ba4ee9c1e4b69fa5f487b2070b78

Here's a Gist that shows the problem:
https://trycf.com/gist/1b7e7b0f1487443b0b59f985bedcf5d0/acf2018?theme=monokai

If you switch to Lucee, `val()` returns an integer.  On ACF, for some reason, `val()` returns a decimal. 
Wrapping `val()` with `int()` should fix the issue like this:
`int( val( generatedKey ) )`

homestar9:
@elpete 
https://github.com/coldbox-modules/quick/pull/201

homestar9:
I just thought of something that might be important for ACF users... what happens if the database is using a `bigint` instead of a regular `int`?  Using `val()` on a massive integer will return notation like this: 9.999999933720369E23.  Wrapping it in `int()` will return the largest possible integer in ACF, which is 9223372036854775807.

ACF Example:
```
writeOutput( int( 999999993372036854775808 ) ); // returns 9223372036854775807 
```

The only thing I can think of right now would be to use Java's `java.math.BigInteger` instead for `bigInt` situations.

Looping in @wpdebruin to see if he has any suggestions.

PR #201, opened 2022-10-14
ACF compatibility: Force generated key to be an integer

homestar9:
See https://github.com/coldbox-modules/quick/issues/200

### DIFF
--- a/models/KeyTypes/AutoIncrementingKeyType.cfc
+++ b/models/KeyTypes/AutoIncrementingKeyType.cfc
@@ -35,7 +35,7 @@ component implements="KeyType" {
 		var generatedKey = arguments.result.result.keyExists( keyName ) ? arguments.result.result[ keyName ] : arguments.result.result.keyExists(
 			"generated_key"
 		) ? arguments.result.result[ "generated_key" ] : arguments.result.result[ "generatedKey" ];
-		arguments.entity.assignAttribute( keyName, val( generatedKey ) );
+		arguments.entity.assignAttribute( keyName, int( val( generatedKey ) ) );
 	}
 
 }


elpete:
Does the PR address your concerns raised [here](https://github.com/coldbox-modules/quick/issues/200#issuecomment-1279347568)?

homestar9:
Yes it does.  However, there's still a problem for those using big integers and ACF.  Both `val()` and `int()` don't work for very big numbers.

elpete:
What is the approach using Java's `BigInteger`?  Could we do that?

homestar9:
> What is the approach using Java's `BigInteger`? Could we do that?

I don't have any experience with `java.math.BigInteger`, but I'll play around with it to see if I can create something that works.  I would be concerned about any possible overhead performance issues, though, since most users will be using normal integers.

homestar9:
`BigInteger` may not be the way to go.   `precisionEvaluate()` may work, but doesn't solve the original intent of using `val()` to always return a number. Therefore, perhaps the better solution is something like this:
```
arguments.entity.assignAttribute( keyName,  ( isNumeric( generatedKey ) ? precisionEvaluate( generatedKey ) : val( generatedKey ) ) );
```

The above code should be tested to ensure it can handle a wide variety of numbers, but the idea here is that it first checks the `generatedKey` to see if it is numeric.  If so, run `precisionEvaluate()` on it to return an actual numeric object. This should fix the value coming back as a string instead.  If the `generatedString` is not a number, `val()` will turn it into 0 (or you can just return 0).

Still, I guess that  `precisionEvaluate()` must come at a performance cost, so if it were me, I would probably set up a new AutoBigIncrementingKeyType.cfc that uses `precisionEvaluate()` to avoid any performance hit for most users.  This PR will still work for most reasonably sized integers.

Issue #186, opened 2022-06-29
Quick does not respect qb's 'returnFormat' setting in coldbox.cfc


ryanalbrecht:
Hi,

In my coldbox.cfc I have the following settings

```	
// quick
quick = {
   defaultGrammar = "MySQLGrammar@qb",
},

// qb
qb = {
   returnFormat: "query"
},
```

If use `.retrieveQuery().get()` on a quick entity, the default return format is not the same as what is set in the coldbox.cfc.

elpete:
Quick maintains it's own qb instance outside of `qb`'s defaults.  This is so it can rely on certain things being true for any queries Quick runs.  One of these things is Quick expects queries to come back as arrays.  If you are wanting to drop down to Quick's internal qb instance and get the result as a query, you are going to need to call `.setReturnFormat( "query" )` on it before executing the query.

This could change in future versions, but it's quite a bit of effort to switch for not too much gain, in my opinion.

elpete:
I would happily accept a PR that handles this, complete with tests.  Otherwise, I probably won't be putting my time towards this.  Thanks for understanding.

PR #206, opened 2022-10-18
Proposal to add ability to set defaultReturnFormat in quick module settings

ryanalbrecht:
in reference to #186 

It seems quick needs to set the underlying qb instance's returnFormat to an `array` to function correctly. I am proposing the following changes to allow a user to set the default return format in the coldbox.cfc module settings. 

The QuickBuilder will now track its own internal returnFormat. When calling `retrieveQuery()` on an entity it will set the return format for the underlying qb instance before returning it to the user. I also implemented a `setReturnFormat()` in the QuickBuilder class so this method call is not forwarded onto the QuickQB instance.    

Thoughts?


ryanalbrecht:
Accidentally change server.json.

Will reopen another pull request

PR #207, opened 2022-10-18
Proposal to add ability to set default ReturnFormat in coldbox modules.

ryanalbrecht:
in reference to #186

It seems quick needs to set the underlying qb instance's returnFormat to an `array` to function correctly. Currently you have to call `setReturnFormat()` on each quick instance if you want to retrieve a query. I am proposing the following changes to allow a user to set the default return format in the coldbox.cfc module settings.

The QuickBuilder will now track its own internal returnFormat. When calling retrieveQuery() on an entity it will set the return format for the underlying qb instance before returning it to the user. I also implemented a setReturnFormat() in the QuickBuilder class so this method call is not forwarded onto the QuickQB instance.

Thoughts?

### DIFF
--- a/ModuleConfig.cfc
+++ b/ModuleConfig.cfc
@@ -9,6 +9,7 @@ component {
 	function configure() {
 		settings = {
 			"defaultGrammar"        : "AutoDiscover@qb",
+			"defaultReturnFormat"	: "array",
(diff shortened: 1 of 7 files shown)


ryanalbrecht:
Strange. All the test are passing on my end for cf2018 and cf2021

PR #209, opened 2022-10-27
Fix Required Parameter Error When Null

homestar9:
When the `guardAgainstNotLoaded()` is disabled, and you attempt to call a relationship, `isNullValue()` will throw the exception: "The required parameter arguments.value was not provided".  Adding a try/catch statement here allows null to pass through without triggering the exception.


bdw429s:

https://community.ortussolutions.com/t/quick-5-what-is-the-optimal-way-to-return-a-subset-of-properties-from-a-relationship/9427/3

ryanalbrecht:
This is related to #213

elpete:
I'm closing this in favor of #213

PR #211, opened 2022-11-10
Fix bad link to getting started

gpickin:


elpete:
docs: fix broken getting started link

PR #213, opened 2023-01-25
Fix a bug when calling isNullValue

ryanalbrecht:
By default when calling `getInstance('entity')` the properties for the given entity will be an actual null value. When retrieve an entity from the database, null fields in the database will be converted to empty strings in the entity.

If you create a new entity and then save it. Any properties that did not have their values set will still be null. This will cause an exception if you then try to retrieve a relationship which has `withDefault()` set. This is caused by cfparam raising an exception if you return null into the 'default' argument.

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -3032,7 +3032,14 @@ component accessors="true" {
 	 * @return  Boolean
 	 */
 	public boolean function isNullValue( required string key, any value ) {
-		param arguments.value = invoke( this, "get" & arguments.key );
+
+		if(!isDefined('arguments.value')){
+			//There is potential for the value of a column to be an actuall null value
+			//We must use isDefined instead of cfparam as returning a null value from invoke
+			//into the 'default' argument of cfparam will raise an exception
+			arguments.value = invoke( this, "get" & arguments.key );
+		}
+
 		if ( isNull( arguments.value ) ) {
 			return true;
 		}


elpete:
Are these two PR's in one then?  I'd also love a test for `clone`. (Also also, could we just do `clone` instead of `cloneEntity`?)

ryanalbrecht:
@elpete yes this is two pull requests in one. The first was a bug fix from a while ago. Im not two sure how to create two separate pull request using git. 

I will implement some tests and rename the function.

ryanalbrecht:
I added some test and also added coverage for the first pull request's bug fix. Also seems like I cannot resolve a conflict on my own commits in this PR so automated tests did not run.

ryanalbrecht:
@elpete a cleaned a bunch of stuff up and this PR should be ready for review and merge

elpete:
Thank you for the code and especially thank you for your patience.

ryanalbrecht:
Thanks elpete, sorry for the messy PRs.

ryanalbrecht:
@elpete quick question. Refering to 'formatting and cleanup' commit you made on this PR. 

Did you manually make those changes or did you use an automated tool?

elpete:
Ha, the PR wasn't messy.  I ran it through CFFormat and changed a couple of the TestBox assertions from `expect( a eq b ).toBeTrue()` to `expect( a ).toBe( b )`.

Issue #212, opened 2022-11-30
regression: where and/or grouping different between v4->v5 for the same scope invocation


davidAtInleague:
We've found some scopes generating bad sql, seems the change happened from the jump to v4->v5.

v4 generates the desired sql.

grammar is mssql.

repro and results we're seeing:
```
/*
writedump(getInstance("RMME_C").withFoo().toSql());

v4.1.3
SELECT [RMME_C].[id_c] FROM [RMME_C] WHERE (([id] = ? AND [id] != ?) OR ([id] = ? AND [id] != ?))

v5.2.0
SELECT [RMME_C].[id_c] FROM [RMME_C] WHERE [id] = ? AND [id] != ? AND [id] = ? AND [id] != ?

v5.2.7
SELECT [RMME_C].[id_c] FROM [RMME_C] WHERE [id] = ? AND [id] != ? AND [id] = ? AND [id] != ?

*/
```
```
component extends="quick.models.BaseEntity" table="RMME_C" accessors=true {
    property name="id_c" type="numeric" sqltype="int";

    variables._key = "id_c";

    function scopeWithFoo(qb) {
        qb.where((X) => {
            X
                .orWhere((X) => X.withBar(1))
                .orWhere((X) => X.withBar(2))
        })
    }
    
    function scopeWithBar(qb, val) {
        qb
            .where("id", "=", val)
            .andWhere("id", "!=", val - 1);
    }
}
```

PR #214, opened 2023-01-26
fixup associativity for nested wheres

davidAtInleague:
`callback(query)` seems like a purely isolated functional thing, but it makes assumptions about its environment, and the existing qb object in the existing environment has where's piled up in it which we don't want to flatten with the nested result. It looks like updating the context (read: the available QB object) during descent into the callback, and then restoring the original context afterwards, keeps the two contexts from flattening their wheres together.

Fixes https://github.com/coldbox-modules/quick/issues/212

### DIFF
--- a/models/QuickQB.cfc
+++ b/models/QuickQB.cfc
@@ -617,4 +617,19 @@ component extends="qb.models.Query.QueryBuilder" accessors="true" {
 		return super.onMissingMethod( argumentCollection = arguments );
 	}
 
+	// override's super impl
+	// keep in sync with `super.whereNested`, or merge them somehow
+	public QueryBuilder function whereNested( required callback, combinator = "and" ) {
+        // We descend into a new QueryBuilder, but it is not an isolated thing and requires a fresh context,
+        // but it will by default inherit the current context.
+        // "relevant context" is assumed to be exactly "the available QueryBuilder object", we can push/pop it.
+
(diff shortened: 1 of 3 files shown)


MordantWastrel:
@elpete We would be happy to have some direction on this as we have one other bug we'd like to submit a PR for, but I know you're working on QB 9 and if you're working on solving any of these differently we won't invest the effort on our side until you're ready for us. We're trying to get Quick 5 merged into our daily workflow ASAP!

elpete:
Just making sure I understand the problem we're solving.

Without this change, do all the where clauses get combined together? Something like this?

```sql
SELECT *
FROM `users`
WHERE (
    `users`.`first_name` = ?
    AND `users`.`last_name` = ?
    OR `users`.`first_name` = ?
    AND `users`.`last_name` = ?
)
```

davidAtInleague:
Hi, yes, we see the following emitted sql, for the same scope invocation across different versions, where 4.1.3 is correct:
```
v4.1.3
SELECT [RMME_C].[id_c] FROM [RMME_C] WHERE (([id] = ? AND [id] != ?) OR ([id] = ? AND [id] != ?))

v5.2.0
SELECT [RMME_C].[id_c] FROM [RMME_C] WHERE [id] = ? AND [id] != ? AND [id] = ? AND [id] != ?
```

PR #215, opened 2023-05-01
ACF2023 and Coldbox 7

jclausen:
Adds tests for ACF2023 and `coldbox@be`


jclausen:
Commit [cb8b5aa](https://github.com/coldbox-modules/quick/pull/215/commits/cb8b5aa9571a0a5e1496a21de5fe8eb5371ed4ae) fixes the Coldbox 7 issues with transient caching

jclausen:
Update:  There is still an issue with this one and Coldbox 7.  The transient cache issues seem to be fixed, but the the test suite, as a whole, just times out and the server becomes unresponsive ( I cancelled the running job after 25 minutes ).

Running each spec individually works fine, but trying to do them all at the same time is an issue.

jclausen:
It looks like @lmajano updated the CI workflow, but this commit still needs to be brought in: https://github.com/coldbox-modules/quick/pull/215/commits/cb8b5aa9571a0a5e1496a21de5fe8eb5371ed4ae

Otherwise you are injecting a transient in to a transient and the cache will make the builder an effective singleton.

PR #219, opened 2023-05-12
Added ability to exclude all global scopes

ryanalbrecht:
As discussed in the slack channel, here is the code + tests to provide ability to exclude all global scopes.


elpete:
Can you re-create the docs PR against 6.1.0?

ryanalbrecht:
> Can you re-create the docs PR against 6.1.0?

Done!

Issue #239, opened 2023-12-07
Add old data to the preupdate event


davidsf:
Currently there are no way to see the old data of the entity in the `preUpdate` event. 

We need this feature in order to audit changes to some entities. 

PR is comming :)

elpete:
Addressed by #293, now merged into .

elpete:
Correction: Addressed by #293, now merged into the next branch.

PR #240, opened 2023-12-07
Add old data to the preupdate event

davidsf:
Send the old data of the entity in the preUpdate event. 

#239

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -1086,6 +1086,7 @@ component accessors="true" {
 				"#parentDefinition.joinColumn#" : parent.keyValues()[ 1 ]
 			} );
 		}
+		var oldData = duplicate(variables._data);
 		guardNoAttributes();
 		guardReadOnly();
 		mergeAttributesFromCastCache();
@@ -1093,7 +1094,7 @@ component accessors="true" {
 		variables._saving = true;
 		var builder       = newQuery();
 		if ( variables._loaded ) {
-			fireEvent( "preUpdate", { entity : this } );
+			fireEvent( "preUpdate", { entity : this, oldData : oldData } );
 			builder
 				.where( function( q ) {
 					arrayZipEach( [ keyNames(), keyValues() ], function( keyName, keyValue ) {


models/BaseEntity.cfc
```diff
@@ -1086,14 +1086,15 @@ component accessors="true" {
...
+		var oldData = duplicate(variables._data);
 		guardNoAttributes();
 		guardReadOnly();
 		mergeAttributesFromCastCache();
 		fireEvent( "preSave", { entity : this } );
 		variables._saving = true;
 		var builder       = newQuery();
 		if ( variables._loaded ) {
-			fireEvent( "preUpdate", { entity : this } );
+			fireEvent( "preUpdate", { entity : this, oldData : oldData } );
```
elpete, line 1097: I’m going to need tests around this, specifically because I don’t think this is correct. `variables._data` here will be the data about to be saved. 

I _think_ you want `variables._originalAttributes` which can also be retrieved calling `get_originalAttributes()` on the entity, which is why nothing extra has been passed in

elpete (changes requested):

models/BaseEntity.cfc
davidsf: You are right, inside the `preUpdate()` I can access  `arguments.eventdata.entity.get_originalAttributes()` and get the orginal data.

Maybe could be document this in [the preUpdate() doc](https://quick.ortusbooks.com/guide/interception-points#quickpreupdate)?  (I think is a common use of preUpdate to access the original data)

elpete: Thinking about it some more, I think it's okay to pass the original attributes to this method, even though it's also available via the entity. 
 Passing the current data might be a good option as well. It's a little unwieldy to get the original attributes right now.  A method like `retrieveOriginalAttributes` or `retrieveOriginalData` would be a good addition as well.

davidsf: Cborm and hibernate sends the old data to `preupdate` event. Just for compatibility with others orm I think it's good to have the same behaviour.

PR #244, opened 2023-12-20
Update HasOneOrMany Relationship Class to Allow Returning a New Entity

homestar9:
Being able to create a new `HasOneOrMany` relationship entity is great, but sometimes you need access to the new entity before persistence.  Adding a `newEntity()` method gives additional flexibility and also plays nicely with other relationship types.

Additionally, being able to call something like `prc.post.postCategories().newEntity()` is much cleaner than using the current workaround, which is `prc.post.postCategories().getRelated().newEntity()`

### DIFF
--- a/models/Relationships/HasOneOrMany.cfc
+++ b/models/Relationships/HasOneOrMany.cfc
@@ -317,9 +317,20 @@ component extends="quick.models.Relationships.BaseRelationship" accessors="true"
 	 * @return      quick.models.BaseEntity
 	 */
 	public any function create( struct attributes = {} ) {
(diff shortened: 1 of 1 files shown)


homestar9:
@elpete I'm not sure why the format test failed, but all other tests appear to have passed. Please let me know if you have any feedback or questions about this PR. It will make a nice addition to entity relationships.

I also updated the docs and submitted a review.

homestar9:
@elpete, I'm sure you're swamped with other projects, so there's no rush on this PR. I just want to check in to make sure you received the notification about this PR and the other pending ones.

PR #259, opened 2025-01-24
Add GUID keyType introduced in QB v9

dbelanger:
This key type is needed for Quick to autogenerate the 36 character GUIDs introduced in QB v9


dbelanger:
If this is approved, you'll need to add the following to the [docs ](https://quick.ortusbooks.com/guide/getting-started/defining-an-entity) under Key Types:

GUIDKeyType

PR #276, opened 2026-05-21
Make the cast decide what null means if exists, else the raw value is returned 

### LINKED ISSUE #275: Should custom casts in Quick be called for NULL database values?
https://community.ortussolutions.com/t/should-custom-casts-in-quick-be-called-for-null-database-values/11099

If a custom cast is defined, the cast should be allowed to decide how NULL is handled.

rodriciru:
Fixes #275

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -617,9 +617,7 @@ component accessors="true" {
 				continue;
 			}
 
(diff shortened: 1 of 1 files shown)


rodriciru:
Any oppinions?

elpete:
Closing as superseded by #279, which implements the same null-cast behavior with broader compatibility handling and regression coverage.

Issue #266, opened 2025-07-15
Calling .get() with options does not pass through the arguments when using quickService


grantcopley:
I'm running into this with Quick v10.0.1 on an app that does not have a default datasource defined in Application.cfc (there are reasons we can't currently define a global datasource).

```
	userService = application.wirebox.getInstance( "quickService:User@someModule" );

	awaitingReview = userService
					.whereNotNull( "Awaiting_Approval_Date" )
					.whereNull( "Approved_Date" )
					.get();
```

I get the following error:

```
Attribute [datasource] is required when attribute [dbtype] is not [query] and no default datasource is defined

you can define a default datasource as attribute [defaultdatasource] of the tag cfapplication or as data member of the Application.cfc (this.defaultdatasource="mydatasource";)
```

The error remains even when I try passing the datasource with the .`get()` call.

```
	awaitingReview = userService
					.whereNotNull( "Awaiting_Approval_Date" )
					.whereNull( "Approved_Date" )
					.get( options = { "datasource": "MyDatasource" } );
```

**Stacktrace:**

```
lucee.runtime.exp.ApplicationException: Attribute [datasource] is required when attribute [dbtype] is not [query] and no default datasource is defined
	at lucee.runtime.tag.Query.doStartTag(Query.java:479)
	at lucee.runtime.functions.query.QueryExecute.call(QueryExecute.java:67)
	at models.grammars.basegrammar_cfc$cf.udfCall1(/qb/models/Grammars/BaseGrammar.cfc:124)
	at models.grammars.basegrammar_cfc$cf.udfCall(/qb/models/Grammars/BaseGrammar.cfc)
	at lucee.runtime.type.UDFImpl.implementation(UDFImpl.java:112)
	at lucee.runtime.type.UDFImpl._call(UDFImpl.java:358)
	at lucee.runtime.type.UDFImpl.callWithNamedValues(UDFImpl.java:213)
	at lucee.runtime.ComponentImpl._call(ComponentImpl.java:699)
	at lucee.runtime.ComponentImpl._call(ComponentImpl.java:594)
	at lucee.runtime.SuperComponent.callWithNamedValues(SuperComponent.java:93)
	at lucee.runtime.util.VariableUtilImpl.callFunctionWithNamedValues(VariableUtilImpl.java:866)
	at lucee.runtime.PageContextImpl.getFunctionWithNamedValues(PageContextImpl.java:1795)
[...]

PR #284, opened 2026-08-22
test: cover QuickService get options

elpete:
## Summary
- add public-API integration coverage for passing query options through a `quickService` chain
- assert the datasource option reaches qb execution through the normal `.get()` flow

Refs #266.

## Reproduction
Current `next` already contains the production fix from `0e0fc11` (`fix(QuickBuilder): Allow passing options for retrieval methods`), so the reported failure does not reproduce unchanged. Temporarily restoring the pre-fix `QuickBuilder.get()` behavior made this regression fail because `options.datasource` was absent; restoring current code made it pass.

## Validation
- `BaseServiceSpec`: 7 passed, 0 failed, 0 errors
- full Lucee 6 suite: 495 passed, 0 failed, 0 errors, 3 skipped
- `box run-script format`
- `git diff --check`

### DIFF
--- a/tests/specs/integration/BaseServiceSpec.cfc
+++ b/tests/specs/integration/BaseServiceSpec.cfc
@@ -1,5 +1,17 @@
 component extends="tests.resources.ModuleIntegrationSpec" {
 
+	function beforeAll() {
+		super.beforeAll();
+		controller
+			.getInterceptorService()
+			.registerInterceptor( interceptorObject = this, interceptorName = "BaseServiceSpec" );
+	}
+
+	function afterAll() {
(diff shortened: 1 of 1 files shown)


elpete:
Reproduction note for #266: I could not reproduce the datasource-option loss on current `next` because commit `0e0fc11` already forwards `get()` arguments through `getEntities()` to qb. I temporarily restored the pre-fix call path locally; the new public-API QuickService test then failed because `options.datasource` never reached `preQBExecute`. Restoring current production code made the test pass. No production change is included here because the behavior is already fixed.

PR #288, opened 2026-08-22
Fix table qualification in whereIn subqueries

### LINKED ISSUE #231: subQuery `.from()` references wrong table (regression from v4 to v6)
We had an entity with the following scope in Quick 4:

```
					q.whereHas( "child", ( q ) => q.whereIn( "familyID", ( q ) => {
						q.from( "family_parents" )
							.where( "clientID", "=", application.clientID )
							.where( "userID", "=", parentUserID )
							.select( "familyID" );
					} ) )
```

which emitted this (focus is on the `whereIn()` portion with `where [clientID] = @P4`

```
WHERE
    [registrations].[clientID] = @P0 AND
    [registrations].[registered] = @P1 AND
    [registrations].[seasonID] = @P2 AND
    [registrations].[childID] != @P3 AND
    EXISTS (
        SELECT 1 FROM [children]
        WHERE
            ([children].[childID] = [registrations].[childID]) AND
            [children].[familyID] IN (
                SELECT [familyID]
                FROM [family_parents]
                WHERE
                    [clientID] = @P4 AND
                    [userID] = @P5
            )
    )
```

In Quick 6/7, it does this:

```
FROM [registrations]
WHERE
    [registrations].[clientID] = @P0 AND
    [registrations].[registered] = @P1 AND
    [registrations].[seasonID] = @P2 AND
    [registrations].[childID] != @P3 AND
    EXISTS (
        SELECT 1 FROM [children]
        WHERE
            ([children].[childID] = [registrations].[childID]) AND
            [children].[familyID] IN (
                -- should be select fp.familyID from family_parents where fp.clientID = x and fp.userID = y
                SELECT [children].[familyID]
                FROM [family_parents]
                WHERE
                    [children].[clientID] = @P4 AND
                    [userID] = @P5
            )
    )
```

This appears to occur only when both the parent object and the subQuery `.from()` table contain the same column: the builder references the parent table's column even with an explicit `.from()`

A workaround is to avoid `.whereIn()` in favor of additional `.whereHas()` but this is a prickly one as it's tough to track down unless you're explicitly testing for valid results in these queries.

### LINKED ISSUE #248: .from in .whereIn subquery incorrectly results in parent entity table
The following Quick scope is producing an erroneous table prefix for the 'where in' clause.
```

     function scopeNotInCloneQueue( query, required string seasonUID ) {
                return query.whereNotIn( "registrationID", function( sQ ) {
                        sQ.from( "registration_clone_queue" )
                                .select( "registrationID" )
                                .join(
                                        "registration_clone_queue_batches",
                                        "registration_clone_queue.batchID",
                                        "=",
                                        "registration_clone_queue_batches.batchID"
                                )
                                .whereTargetSeasonUID( seasonUID )
                } );
        }
```

### What did you expect to happen?

The subquery should be selecting registrationID from registration_clone_queue.

### What actually happened instead?

The SQL produced was to SELECT registrationID FROM registrations (the table of the parent entity where the scope lives). registrationID is the primary key.

A workaround is to specify an alias in the `from` clause and refer to the registraitonID explicitly with the table prefix.

### Environment

List the software versions you're using:

 - Quick: 7.3.1 / Lucee 5

elpete:
Closes #248
Closes #231

## Issue review

Issues #248 and #231 are the same reproducible table-qualification bug at different nesting depths. A closure-based `whereIn`/`whereNotIn` subquery inherits Quick's entity-aware column formatter. After the closure changes the subquery table with `from()`, attribute names shared with the parent entity were still qualified against the parent table.

Recommendation: **10/10 - implement.**

Reasons for:

- the generated SQL can silently return incorrect records, not merely throw an error;
- closure subqueries and relationship constraints are supported public qb/Quick APIs;
- the fix retains attribute-to-column mapping while using the subquery's actual table or alias;
- the same correction covers direct subqueries (#248) and subqueries nested inside `whereHas` (#231).

Tradeoffs:

- subquery columns are now explicitly qualified with their current table/alias, which can change exact SQL-string snapshots;
- the formatter adjustment is deliberately scoped to closure-based `whereIn`/`whereNotIn` subqueries rather than changing qualification globally.

## Reproduction and fix

The #248 regression uses a direct public `whereNotIn` query. Before the fix, current `next` with qb 14.0.0-beta.3 generated:

```sql
SELECT `users`.`id` FROM `countries`
```

The #231 regression recreates the reported nested public flow:

```cfml
getInstance( "Registration" ).whereHas( "child", function( q ) {
    q.whereIn( "familyID", function( subquery ) {
        subquery.from( "family_parents" ).select( "familyID" );
    } );
} );
```

Before the fix it generated:

```sql
SELECT `children`.`familyID` FROM `family_parents`
```

and could return a false-positive registration. The callback now installs a formatter tied to the subquery builder, producing `countries.id` and `family_parents.familyID` respectively. Table aliases are also honored.

## Validation

- #248 test-first reproduction: 9 passed, 1 failed, 0 errors before implementation.
- #231 test-first reproduction: 37 passed, 1 failed, 0 errors before implementation.
- focused relationship-query suite after fix: 38 passed, 0 failed, 0 errors.
- full Lucee 6 suite using qb 14.0.0-beta.3: 497 passed, 0 failed, 0 errors, 3 skipped.
- `box run-script format`
- `git diff --check`

### DIFF
--- a/models/QuickBuilder.cfc
+++ b/models/QuickBuilder.cfc
@@ -862,6 +862,24 @@ component accessors="true" transientCache="false" {
 	}
 
 	/**
+	 * Qualifies a column using the current table for the provided query.
+	 *
+	 * @column The column to qualify.
+	 * @query  The query whose current table should be used.
+	 *
+	 * @return string
+	 */
+	public string function qualifyColumnForQuery( required string column, required any query ) {
+		var tableName = arguments.query.getTableName();
+		if ( arguments.query.getAlias() != "" ) {
+			tableName = arguments.query.getAlias();
+		}
+		return !isNull( tableName ) && isSimpleValue( tableName ) && tableName != ""
+		 ? qualifyColumn( arguments.column, tableName )
+		 : qualifyColumn( arguments.column );
+	}
+
+	/**
 	 * Returns the table name of the underlying entity
 	 */
 	public string function tableName() {
--- a/models/QuickQB.cfc
+++ b/models/QuickQB.cfc
@@ -80,6 +80,15 @@ component
 	) {
 		if ( isStruct( arguments.values ) && structKeyExists( arguments.values, "isQuickBuilder" ) ) {
 			arguments.values = arguments.values.getQB();
+		} else if ( isClosure( arguments.values ) || isCustomFunction( arguments.values ) ) {
+			var callback     = arguments.values;
+			var quickBuilder = variables.quickBuilder;
+			arguments.values = function( query ) {
+				query.setColumnFormatter( function( column ) {
+					return quickBuilder.qualifyColumnForQuery( column, query );
+				} );
+				return callback( query );
+			};
 		}
 
 		return super.whereIn( argumentCollection = arguments );
--- a/tests/specs/integration/BaseEntity/Relationships/QueryingRelationshipsSpec.cfc
+++ b/tests/specs/integration/BaseEntity/Relationships/QueryingRelationshipsSpec.cfc
@@ -38,6 +38,23 @@ component extends="tests.resources.ModuleIntegrationSpec" {
 						expect( users ).toHaveLength( 1 );
 					} );
 
+					it( "uses a whereIn subquery table inside whereHas", function() {
+						queryExecute(
+							"INSERT INTO family_parents ( parentID, familyID ) VALUES ( :parentID, :familyID )",
+							{ "parentID" : 1, "familyID" : 2 },
+							{ "datasource" : "quick" }
+						);
+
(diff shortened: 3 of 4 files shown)


elpete:
## Additional issue review: #231

Addresses #231. **Fit score: 10/10.** This is the same silent table-qualification regression covered by this PR. Supporting explicit `from()` tables inside `whereIn()` is essential because returning plausible but incorrect records is worse than a hard failure. The branch includes the exact public `whereHas` + `whereIn` regression first, and the shared fix correctly qualifies the nested select from `family_parents`. The tradeoff is limited to correcting SQL ownership; no public API changes.

Issue #245, opened 2024-01-25
[GitBook] Create() Method Documentation Improvement


homestar9:
@elpete, I no longer have access to the Quick documentation on Gitbook for some reason, so I'm posting my suggestion here:

On the page: https://quick.ortusbooks.com/guide/getting-started/creating-new-entities#create
It is critically important to let users know the `create()` method returns a brand new entity, and does not persist the current entity which may have called it.

Example:
```
prc.user = getInstance( "User" ).newEntity();

// danger Will Robinson! This will not update prc.user.
prc.user.create( {
 "firstName": "Dave"
} );

writeDump( prc.user.isLoaded() ); // <--- WILL BE FALSE!
```

If you want to use create, you must do this instead:
```
prc.user = getInstance( "User" ).newEntity();

// in order for this to work, you must replace the original variable.
prc.user = prc.user.create( {
 "firstName": "Dave"
} );

writeDump( prc.user.isLoaded() ); // <--- WILL BE TRUE
```

elpete:
The create() page now explicitly warns that create is a factory returning a new saved instance and shows assigning its return value.

[Documentation](https://github.com/ortus-docs/quick-docs/blob/13.0.0/guide/getting-started/creating-new-entities.md). Merged into the current 13.0.0 documentation branch via [the documentation PR](https://github.com/ortus-docs/quick-docs/pull/64).

Closing with the supported approach documented.

PR #290, opened 2026-08-22
Clarify that create returns a separate entity

elpete:
Addresses #245

## Issue review

This is a useful and accurate documentation improvement. `BaseEntity.create()` is a factory-style convenience method: it calls `newEntity()`, fills that new instance, saves it, and returns it. Calling `create()` does not mutate or load the receiver. That distinction is easy to miss and can leave application variables pointing at an unloaded entity.

Recommendation: **9/10 — document explicitly.**

Reasons for:
- misunderstanding the return value can produce subtle application bugs
- the implementation has always made the separate-instance behavior explicit, but the user-facing wording does not
- a short assignment example makes the safe usage obvious

Tradeoffs:
- the GitBook page itself is maintained in the separate `ortus-docs/quick-docs` repository on version branches; that repository has no `next` branch
- this PR therefore updates Quick’s source/API documentation and locks the behavior down in `next`, while the same wording should be carried to the versioned GitBook source when documentation is published

## Implementation

The `BaseEntity.create()` documentation now states that it always creates and returns a separate instance, does not mutate the receiver, and shows assigning the return value.

A public API integration test demonstrates the exact concern from the issue:
- the original entity remains unloaded and has no username
- the returned entity is loaded and has the created username

No runtime behavior was changed.

## Validation

- focused CreateSpec: 4 passed, 0 failed, 0 errors
- full Lucee 6 suite using qb 14.0.0-beta.3: 496 passed, 0 failed, 0 errors, 3 skipped
- formatter completed
- `git diff --check` passed

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -1317,6 +1317,16 @@ component accessors="true" {
 	/**
 	 * Creates a new entity with the given attributes and then saves the entity.
 	 *
+	 * This method always creates and returns a separate entity instance. It does
+	 * not fill, save, or otherwise mutate the entity on which it was called.
+	 * Assign the returned entity when it needs to be used after creation:
+	 *
+	 * ```
+	 * var user = getInstance( "User" ).newEntity();
+	 * user = user.create( { "firstName" : "Dave" } );
+	 * user.isLoaded(); // true
+	 * ```
+	 *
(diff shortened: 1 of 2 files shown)


elpete:
Moved to the canonical Quick documentation repository as ortus-docs/quick-docs#64.

PR #294, opened 2026-08-22
fix: advance nested has relationship paths

### LINKED ISSUE #153: Stack overflow for `has` method when going through `BelongsToMany` -> `BelongsToMany` -> `hasMany`

### LINKED ISSUE #238: Connected entities fail if more than 2 are connect with hasManyThrough Error building: HasMany@quick -> java.lang.StackOverflowError
Please see: https://github.com/akitogo/quicktest

```
getInstance("user")
            .with("usergroup.permission.extendedpermission")
            .has("usergroup.permission.extendedpermission")
```
and I use has, it works with two like `has("usergroup.permission")`  but as soon I have three (see above) I get (bearbeitet) 
an `Error building: HasMany@quick -> java.lang.StackOverflowError`

elpete:
## Issue review

Issues #238 and #153 both report stack overflows from public `has()` calls across relationship paths longer than two segments. #153 specifically traverses `BelongsToMany -> BelongsToMany -> hasMany`.

I rate both **10/10** for Quick. Arbitrarily deep relationship paths are part of the public `has()` contract, and a stack overflow is severe and difficult for callers to diagnose. The main implementation concern is preserving each relationship type's specialized comparison and pivot-table behavior while composing the nested query.

## Reproduction

I recreated both reports with public Quick entity flows:

- #238: three nested `hasManyThrough` relationships.
- #153: `Post.has("tags.posts.comments")`, which traverses two `BelongsToMany` relationships followed by a polymorphic `hasMany`.

On current `next` with `qb 14.0.0-beta.3`, both tests reproduced a `StackOverflowError`. The #153 regression overflowed after about 35 seconds.

## Root cause and fix

The old `hasNested()` recursion did not advance the remaining relationship path, which recursively resolved the same relationship forever.

An initial path-advancement repair stopped the overflow but generated invalid SQL for #153 by losing the enclosing `tags` query and referencing `tags.id` outside its scope. The final implementation composes nested `has()` calls through the established `whereHas()` relationship path. That lets each relationship type build and nest its own comparison constraints, including `BelongsToMany` pivot queries, and removes the duplicate recursive implementation.

Both regression tests use public Quick builder/entity APIs and execute the generated queries.

## Validation

- Before fix: focused suite produced 2 `StackOverflowError` errors.
- After fix: focused relationship-query suite: 39 passed, 0 failed, 0 errors.
- Full suite: 497 passed, 0 failed, 0 errors, 3 skipped.
- `box run-script format`
- `git diff --check`

Closes #238
Closes #153

### DIFF
--- a/models/QuickQB.cfc
+++ b/models/QuickQB.cfc
@@ -580,14 +580,19 @@ component
 			q = q.getQB();
 		}
 
-		arguments.relationQuery = invoke(
-			arguments.relationQuery,
-			"whereExists",
-			{ "query" : q }
-		);
+		var parentQuery            = arguments.relationQuery;
+		arguments.relationQuery    = q;
+		arguments.relationshipName = listRest( arguments.relationshipName, "." );
+		var nestedQuery            = hasNested( argumentCollection = arguments );
+		if ( structKeyExists( nestedQuery, "getQB" ) ) {
+			nestedQuery = nestedQuery.getQB();
+		}
 
-		var result = hasNested( argumentCollection = arguments );
-		return structKeyExists( result, "getQB" ) ? result.getQB() : result;
+		return invoke(
+			parentQuery,
+			"whereExists",
+			{ "query" : nestedQuery }
+		);
 	}
 
 	/**
--- a/models/QuickQB.cfc
+++ b/models/QuickQB.cfc
@@ -411,6 +411,25 @@ component
 		string combinator = "and",
 		boolean negate    = false
 	) {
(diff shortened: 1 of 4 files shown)


elpete:
## Additional issue review: #153

Addresses #153. **Fit score: 9/10.** Deep `has()` paths must compose without recursive builder state causing stack overflow. This branch includes the nested BelongsToMany -> BelongsToMany -> HasMany regression and replaces the recursive composition with bounded nested-query handling. The main tradeoff is internal complexity, covered by the relationship-query suite.

PR #299, opened 2026-08-22
fix: refresh scoped virtual projections

### LINKED ISSUE #169: fresh/refresh don't update subselect
I've got a scope like 
```
function scopeWithFoo( qb ) {
    return qb.addSubselect("foo", newEntity("someEntity").select("foo").whereSomething());
}
```
later we use the scope, but calling fresh/refresh doesn't update `foo`:
```
var base = getInstance("theModel").withFoo().firstOrFail();

// ... do things that update the foo value indirectly

var refreshedBase = base.fresh(); 
// does not reflect the change
refreshedBase.getFoo();

// does not reflect the change
base.refresh().getFoo(); 

// can see the change like this
getInstance("theModel").withFoo().whereTheModelID(base.getID()).getFoo();
```
thanks!

### LINKED ISSUE #220: refresh does not reload scopes
```
component table="RMME_A" extends="quick.models.BaseEntity" accessors=true {
    property name="id_a" type="numeric" sqltype="int";
    property name="v" type="string" sqltype="varchar";
    
    variables._key = "id_a";
    function keyType() {
		return variables._wirebox.getInstance( "NullKeyType@quick" ); // the key comes from Stripe and is not generated locally
	}

    function scopeWithComputedValue( qb ) {
        qb.appendVirtualAttribute( "computedValue" )
        qb.selectRaw("'the value in the [v] column is: <<<' + [v] + '>>>' as computedValue")
    }
}
///////
queryExecute("
	drop table if exists rmme_a;
	create table rmme_a(id_a int, v varchar(max));
")

getInstance("RMME_A").create({id_a:1, v: "AAA"})

x = getInstance("RMME_A").whereID_A(1).withComputedValue().firstOrFail()
writedump(x.getComputedValue()) // "the value in the [v] column is: <<<AAA>>>"
x.update({v: "BBB"})
x = x.refresh()
writedump(x.getV()) // "BBB", ok
writedump(x.getComputedValue()) // "the value in the [v] column is: <<<AAA>>>", computed value out of sync with actual value after a refresh


x = getInstance("RMME_A").whereID_A(1).withComputedValue().firstOrFail()
writedump(x.getComputedValue()) // "the value in the [v] column is: <<<BBB>>>", ok after full firstOrFail reload

abort;
```

This can lead to entities winding up in inconsistent states, where computed values from scopes are no longer in sync with actually refreshed data.

This behavior appears unchanged from Quick4, maybe it just needs to be documented.

elpete:
Closes #220
Closes #169

## Issue review

**Recommendation: 9/10.** Both reports describe the same correctness boundary: fresh or refreshed entities should not retain or discard derived values when the persisted data is reloaded. Replaying the projection that produced virtual data is a strong fit because `fresh()` and `refresh()` promise a current database representation.

Reasons for:
- Stale computed values leave one entity containing mutually inconsistent attributes.
- Dropped subselects make a fresh entity structurally different from the entity it replaces.
- Callers should not need to know whether a value came from a global scope, local scope, or explicit subselect.
- The originating qb query is the authoritative definition of raw and derived projections.

Tradeoffs:
- Entities containing virtual query data retain a cloned query snapshot until discarded.
- Replaying a scoped query preserves its original constraints; if the row no longer matches, the query cannot retrieve it, consistent with the scoped view that produced the entity.

## Reproduction

For #220, a public entity flow using `withFullName()` reproduced a stale virtual value after `refresh()`: the persisted attribute changed but the derived projection did not.

For #169, a public entity flow using `withLatestPostId()` reproduced the dropped subselect after `fresh()`. After changing which post was latest, the regression test expected the new post id but received an empty value. The focused bundle had 35 passes and 1 failure before implementation.

## Implementation

- Capture the underlying qb query when a loaded result contains virtual attributes.
- Replay that snapshot during both `fresh()` and `refresh()`, constrained by the entity key.
- Rehydrate raw replay results from `fresh()` into a normal Quick entity and retain the snapshot for later refreshes.
- Keep the existing path for entities without virtual query data.
- Add public integration regressions covering scoped projections and subselects.

## Validation

- Focused GetSpec: 36 passed, 0 failed, 0 errors
- Adjacent Subqueries/GlobalScope/ChildClass bundles: 48 passed, 0 failed, 0 errors, 1 skipped
- Full suite: 497 passed, 0 failed, 0 errors, 3 skipped
- `box run-script format`
- `git diff --check`

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -1067,13 +1067,19 @@ component accessors="true" {
 	 * @return  quick.models.BaseEntity
 	 */
 	public any function fresh() {
-		return newQuery()
+		var freshEntity = isNull( variables._refreshQuery ) ? newQuery() : variables._refreshQuery.clone().offset( 0 );
+		var freshData   = freshEntity
+			.from( tableName() )
 			.where( function( q ) {
 				arrayZipEach( [ keyNames(), keyValues() ], function( keyName, keyValue ) {
-					q.where( keyName, keyValue );
+					q.where( this.qualifyColumn( keyName ), keyValue );
 				} );
 			} )
 			.first();
+		if ( !isStruct( freshData ) || structKeyExists( freshData, "isQuickEntity" ) ) {
+			return freshData;
+		}
+		return newEntity().hydrate( freshData ).set_refreshQuery( variables._refreshQuery );
 	}
 
 	/**
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -200,6 +200,12 @@ component accessors="true" {
 	 */
 	property name="_virtualAttributes" persistent="false";
 
+	/**
+	 * A snapshot of the query used to load this entity. It is replayed by refresh
+	 * so scoped projections and other query customizations stay in sync.
+	 */
+	property name="_refreshQuery" persistent="false";
+
 
 	/**
 	 * A boolean flag indicating that the entity has been loaded from the database.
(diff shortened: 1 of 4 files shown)


elpete:
## Additional issue review: #169

Addresses #169. **Fit score: 9/10.** `fresh()` and `refresh()` should retain scoped virtual projections because those projections are part of the entity shape the caller requested. This branch contains a public regression and the implementation that preserves subselects when rebuilding the lookup. The only tradeoff is that deliberately dropping a projection now requires a new unscoped query.

PR #301, opened 2026-08-22
fix: synchronize loaded has-many relationships

### LINKED ISSUE #124: .deleteAll() substitutes foreign key with "true"
We have an `invoice` entity that `hasMany` `lineItems`, and in the `preDelete()` interceptor we had this code:

`this.lineItems().deleteAll()`

The resulting query was (from memory now, but the important part is right)

`DELETE FROM invoice_lineItems WHERE invoiceID = [true]`

`this.getLineItems()` works fine, as does `this.ineItems().toSQL()` -- the FK is correctly populated.

As a workaround, this works:

```
this.getLineItems()
	.each( ( l ) => {
		l.delete();
	} );
```

Overkill for what's needed but we only ever have 1-3 line items so it's not a big deal. I don't think I've ever used .deleteAll() before now. 

Quick 4.1.3, qb 8.0.3

### LINKED ISSUE #125: .deleteAll() should dissociate from the parent
Separately from https://github.com/coldbox-modules/quick/issues/124 - if on `User` I call `posts().deleteAll()`, it seems at the moment that it gets passed to QB but doesn't actually change the in-memory relationships on `User`. Should it not clear those out as well?

### LINKED ISSUE #149: Improve relationship assignment when creating new entities off of relationships
Here's two examples for `BelongsTo` and `HasMany`: (other relationships would need to be thought through, as well)

When creating an entity off of a `BelongsTo` relationship, the new related entity should be assigned to the relationship for the child entity.  For example:

```cfc
// Post.cfc
component extends="quick.models.BaseEntity" accessors="true" {

    function author() {
        return belongsTo( "User" );
    }

}
```
```cfc
var post = getInstance( "Post" ).findOrFail( 1 );
var newAuthor = post.author().create( { ... } );
post.isRelationshipLoaded( "author" ) == true;
post.getAuthor() == newAuthor;
```

The same should be true for the child entity when creating it through a `HasMany` relationship:

```cfc
// User.cfc
component extends="quick.models.BaseEntity" accessors="true" {

    function posts() {
        return hasMany( "Post" );
    }

}
```
```cfc
var user = getInstance( "User" ).findOrFail( 1 );
var newPost = user.posts().create( { ... } );
newPost.isRelationshipLoaded( "author" ) == true;
newPost.getAuthor() == user;
```

It gets tricky when trying to handle the `many` side of the relationship.  The parent entity may have other child entities related to it, and it would be incorrect to assign the relationship with only the newly created record.  Even in the case of creating a new parent entity, it is _possible_ (though unlikely) that the newly created parent entity already has related child entities.

One option is to only append the new record when the relationship is loaded.  Loading a relationship is accepting that the data was current at that point, so we can safely add a new record without worry.  Another option is to give some way for a developer to initialize the relationship as blank when creating.

One last thing to consider is if we add this feature should there be a way to disable it?  I can't think of any use case that this would be a problem with, but it's at least worth asking the question.

### LINKED ISSUE #190: Setting HasMany() Relationship After Relationship Loaded Fails to Update Relationship
If you update an entity's `hasMany` relationship _after_ the relationship has been loaded and then attempt to retrieve the relationship, it will not reflect the changes unless you `refresh()` the entity first.

This is problematic in situations where you need to validate the data you assigned to the entity either via `fill`, `populateModel`, `saveMany`, or `set` methods.  The only workaround is to reload the entity from the database first, which could introduce performance problems when dealing with many operations. It also creates inconsistent behavior where some relationships will reflect the changes and others will not after setting the data.

I suspect the issue has to do with the `_relationshipsLoaded` property being set to `true` after you initially retrieve the data and Quick not updating  `_relationshipsData` when the setter is called (or however the data was set).

The solution IMO would be to automatically force set `_relationshipsData` any time the relationship data is updated via setter, fill, saveMany(), etc.  This would prevent unneeded trips to the database and make relationship setting/getting more consistent across relationship types.

Code sample of the problem:
```
// create or get an entity (creating is easier for sake of example)
prc.task = getInstance( "Task" ).create(); 

// Load the relationship (note: you can also just eager load the relationships above)
prc.task.getTaskLocations();

// sample
prc.locations = [
    getInstance( "TaskLocation" ).create(), 
    getInstance( "TaskLocation" ).create()
];

// Now set the relationship. You have a couple options here:

// 1. Setter method (default for sake of example)
prc.task.setTaskLocations( prc.locations );

// 2. Fill method
//prc.task.fill( { taskLocations: prc.locations } );

// 3. Save Many
//prc.task.taskLocations().saveMany( prc.locations ); 

// now try to access the updated relationsihp.  It will be empty.
// you won't be able to validate against it, or do anything with the relationship data.
writeDump( var=prc.task.getTaskLocations(), top=2 ); // dump the relationship data.

prc.task.refresh(); // refresh the relationship. Danger: This will erase any other data you set but did not save() yet!

// Now the relationship data is loaded because a fresh copy was created from the database.
writeDump( var=prc.task.getTaskLocations(), top=2 ); // dump the relationship data.
```

elpete:
Closes #190
Closes #149
Closes #125
Closes #124

### #190 / #125 — synchronize loaded has-many relationships

**Recommendation: 9/10.** Mutating a relationship should update an already-loaded relationship cache. Returning stale data until `refresh()` is incorrect and unnecessarily expensive.

### #149 — assign relationships when creating through relationships

**Recommendation: 8/10.** Created related entities should expose relationship state Quick already knows, without an immediate query.

### #124 — preserve relationship constraints in `deleteAll`

**Recommendation: 10/10.** A relationship deletion must keep its foreign-key constraint and bindings. Current Quick/qb already does; the new public regression protects the executed result.

Reasons for:

- `BelongsTo.create()` knows the created parent and can safely associate/cache it.
- A loaded has-many collection represents an accepted snapshot; relationship mutations should keep that snapshot coherent.
- Seeding the inverse before saving makes it available to lifecycle events.
- Owning `deleteAll()` at the relationship layer allows Quick to synchronize or invalidate the parent cache after the database mutation.

Tradeoffs:

- Quick cannot safely infer an inverse alias such as `author` from `User.posts()`; the inverse relationship name is therefore optional and explicit.
- An unloaded parent collection is deliberately left unloaded because other related rows may already exist.
- `BelongsTo.create()` updates the child foreign key in memory but does not silently save the existing child.
- An ID-restricted relationship deletion invalidates a loaded cache instead of partially filtering composite identifiers in memory; the next getter reloads authoritative state.

## Reproduction

For #190, direct `saveMany()` reproduced stale loaded data: after loading two posts and saving two more, `getPosts()` still returned two.

For #149, public regressions showed:

- `post.author().create(...)` was forwarded through the builder and errored because no relationship-level `create()` existed.
- `user.posts().create(..., "author")` created the post but left its inverse relationship unloaded and required a refreshed parent to see it.

For #125, a public regression loaded two posts, called `user.posts().deleteAll()`, and showed that the database rows were deleted while `user.getPosts()` still returned the stale two-item cache.

For #124, the historical `true` foreign-key binding no longer reproduces with qb 14.0.0-beta.3: the same regression deletes exactly two constrained posts and a fresh parent returns an empty collection.

All tests use public Quick entity and relationship APIs.

## Implementation

- Keep loaded has-many caches synchronized after `saveMany()` and replacement setters.
- Add `BelongsTo.create()` to create the parent, associate it to the existing child, and cache the relationship.
- Extend has-one/has-many `create()` with an optional inverse relationship name, seeded before save.
- Append a created child to an already-loaded collection, while leaving an unloaded collection untouched.
- Reject unknown inverse relationship names before saving.
- Add relationship-owned `deleteAll()` that clears a fully deleted loaded relationship and invalidates ID-restricted loaded caches.

## Validation

- Focused `HasManySpec`: 17 passed, 0 failed, 0 errors.
- Full suite: 500 passed, 0 failed, 0 errors, 3 skipped.
- `box run-script format`
- `git diff --check`

### DIFF
--- a/models/Relationships/BelongsTo.cfc
+++ b/models/Relationships/BelongsTo.cfc
@@ -330,6 +330,23 @@ component
 	}
 
 	/**
+	 * Creates the related parent entity, associates it to the child, and caches
+	 * it as the loaded relationship value.  The child entity is not saved.
+	 *
+	 * @attributes  The attributes for the new related entity.
+	 *
+	 * @return      quick.models.BaseEntity
+	 */
+	public any function create( struct attributes = {} ) {
+		var createdEntity = variables.related
+			.newEntity()
+			.fill( arguments.attributes )
+			.save();
+		associate( createdEntity );
+		return createdEntity;
+	}
+
+	/**
 	 * Removes an entity as the parent of the relationship.
 	 * For example, if a Post belongs to a User, dissociate will set the
 	 * foreign key column on the Post entity to null.
--- a/models/Relationships/HasOneOrMany.cfc
+++ b/models/Relationships/HasOneOrMany.cfc
@@ -272,7 +272,9 @@ component
 			}, {} ),
 			force = true
 		);
-		return saveMany( argumentCollection = arguments );
+		var savedEntities = saveMany( argumentCollection = arguments );
+		variables.parent.assignRelationship( variables.relationMethodName, savedEntities );
+		return savedEntities;
 	}
 
 	/**
@@ -284,11 +286,20 @@ component
 	 * @return        [quick.models.BaseEntity]
 	 */
 	public array function saveMany( required any entities ) {
-		arguments.entities = isArray( arguments.entities ) ? arguments.entities : [ arguments.entities ];
+		arguments.entities        = isArray( arguments.entities ) ? arguments.entities : [ arguments.entities ];
+		var relationshipWasLoaded = variables.parent.isRelationshipLoaded( variables.relationMethodName );
+		var loadedEntities        = relationshipWasLoaded ? variables.parent.retrieveRelationship(
+			variables.relationMethodName
+		) : [];
 
-		return arguments.entities.map( function( entity ) {
+		var savedEntities = arguments.entities.map( function( entity ) {
 			return save( arguments.entity );
 		} );
+		if ( relationshipWasLoaded ) {
+			loadedEntities.append( savedEntities, true );
+			variables.parent.assignRelationship( variables.relationMethodName, loadedEntities );
+		}
+		return savedEntities;
 	}
 
 	/**
--- a/models/Relationships/HasOneOrMany.cfc
+++ b/models/Relationships/HasOneOrMany.cfc
@@ -303,6 +303,31 @@ component
 	}
 
 	/**
+	 * Deletes entities matching the relationship query and synchronizes a loaded parent cache.
+	 *
+	 * @ids  An optional array of related entity ids to delete.
+	 *
+	 * @return  { "query": QueryBuilder Return Format, "result": struct }
+	 */
+	public struct function deleteAll( array ids = [] ) {
+		var result = variables.relationshipBuilder.deleteAll( arguments.ids );
+
+		if ( variables.parent.isRelationshipLoaded( variables.relationMethodName ) ) {
+			if ( arguments.ids.isEmpty() ) {
+				var loadedValue = variables.parent.retrieveRelationship( variables.relationMethodName );
+				variables.parent.assignRelationship(
+					variables.relationMethodName,
+					isArray( loadedValue ) ? [] : javacast( "null", "" )
+				);
+			} else {
+				variables.parent.clearRelationship( variables.relationMethodName );
+			}
+		}
+
+		return result;
+	}
+
+	/**
 	 * Associates an entity or key value for an entity to the parent entity.
 	 *
 	 * @entity   An entity or key value for an entity to associate.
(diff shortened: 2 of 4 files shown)


elpete:
For #124, I attempted the original relationship `deleteAll()` reproduction against current Quick and qb 14.0.0-beta.3. The historical `foreignKey = true` binding no longer occurs: the public regression deletes exactly the two related rows and a fresh parent confirms the collection is empty. The same test reproduced #125’s stale loaded cache, which this branch now fixes. Both issue analyses and closing references are in the PR body.

elpete:
## Additional issue reviews: #149 and #125

Addresses #149. **Fit score: 8/10.** Created related entities should update relationship state where Quick can do so without pretending an unloaded collection is complete. This branch assigns the inverse relation and appends to an already-loaded has-many cache; the tradeoff is intentionally leaving unloaded collections unloaded.

Addresses #125. **Fit score: 8/10.** `deleteAll()` should clear an already-loaded in-memory relationship after the database delete. This branch synchronizes that cache. It does not eagerly load an unloaded relationship merely to clear it, avoiding unnecessary queries.

PR #304, opened 2026-08-22
feat: return empty relationships for new entities

### LINKED ISSUE #131: Return nulls instead of throwing an error when attempting to access a relationship on an unloaded entity
Quick's behavior is to throw a hard error when attempting to access a relational property on an unloaded entity. 

It  is often common to perform post-fill checks on properties to set them to their defaults. For example:

```
user.new( rc );
if( isNull( user.getCreatedBy() ) ){  // will throw a hard error
   user.setCreatedBy( getCurrentUser() );
}
```

Suggestion would be to add a module setting or change the default behavior to simply return null when the relationship is requested and it has not been associated.

This also promotes better ease of handling entity CRUD interception points, by not requiring knowledge of the filled values.

### LINKED ISSUE #148: Improve `retrieveRelationship`
+ [ ] Relationships that don't exist should throw instead of returning `null`.
+ [ ] Return relationship's default value when missing from `retrieveRelationship`.
+ [ ] Allow passing a default value?

### LINKED ISSUE #177: Idea/Improvement: Return Null When Requesting Relationship Data from Non Loaded Entities
Related to [this post](https://community.ortussolutions.com/t/quick-4-2-4-this-instance-is-not-loaded-so-it-cannot-access-the-relationship/9216) on the Ortus community forums.

It would be great if, when requesting relationship data from a non-loaded (new) entity, Quick would return null (or an empty array where appropriate) just like if you performed the same operation from a loaded entity. This behavior should be more intuitive and also better follows the null object pattern.

Here's a simple example of how one might use this new change in Quick:
```
// Post.cfc
component 
    extends="quick.models.BaseEntity" 
    accessors="true" 
{

    // Post belongs to an Author
    function author() {
        return belongsTo( "User" );
    }

}

// Posts.cfc Handler
function index( event, rc, prc ) {
	
	// Get an existing entity, or return a new one if not found
	prc.post = getInstance( "Post" )
		.firstOrCreate( rc.id );

	// return the memento with the author (it will either be null or populated)
	return prc.post.getMemento( "author" );

}
```

elpete:
Closes #177
Closes #148
Closes #131

### #177 / #131 — empty relationship values for new entities

**Recommendation: 9/10.** New entities should expose a relationship type's normal empty value instead of forcing loaded-state branching in forms, post-fill defaults, interceptors, and mementos. This should be the default rather than an opt-in setting so a relationship getter has the same value shape on new and loaded entities.

### #148 — improve `retrieveRelationship()`

**Recommendation: 9/10.** The explicit relationship-cache API should validate names and provide predictable defaults.

Reasons in favor:

- to-one relationships consistently return `null` or a configured default entity;
- to-many relationships consistently return an empty array;
- callers can seed an explicit default without using internal state;
- invalid names fail close to the call site with a useful exception;
- no meaningless SQL is issued;
- post-fill code can inspect a new entity's relationships without knowing which values were supplied.

Tradeoffs:

- retrieving an unloaded relationship through `retrieveRelationship()` now marks its default value loaded, so callers choosing this low-level cache API are intentionally suppressing a later lazy-load query;
- relationship definitions are resolved to validate them, so errors inside a real relationship method continue to propagate rather than being mislabeled as a missing relationship;
- direct relationship-builder calls on new entities still throw because querying or mutating from an unpersisted parent remains invalid.

## Reproduction

Public API tests reproduced the prior behavior:

- getters on new entities threw `QuickEntityNotLoaded`, including the #131 post-fill pattern;
- `retrieveRelationship("posts")` returned null instead of an empty collection;
- an explicit default argument was ignored;
- an unknown relationship returned null instead of throwing.

Before the #148 implementation, its focused bundle reported 9 passed, 1 failure, and 2 errors.

## Implementation

- Initialize new-entity getters through each relationship's existing `initRelation()` behavior.
- Add a base null initializer for to-one relationships, including polymorphic belongs-to.
- Validate relationship names in `retrieveRelationship()`.
- Return and cache the relationship type default without querying when no value is loaded.
- Accept an optional caller-provided default value and cache it.
- Preserve configured `withDefault()` entities and empty arrays for collection relationships.
- Keep loaded cached retrieval on the fast path without rebuilding relationship objects.

## Validation

- Focused relationship/default/polymorphic suites: 16 passed, 0 failed, 0 errors.
- Full suite: 500 passed, 0 failed, 0 errors, 3 skipped.
- `box run-script format`
- `git diff --check`
- Dependency baseline: `qb 14.0.0-beta.3` (`qb@be`).

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -1919,10 +1919,7 @@ component accessors="true" {
 	 * @return  quick.models.Relationships.BaseRelationship
 	 */
 	private any function resolveRelationship( required string name ) {
-		var relationshipName = arguments.name;
-		var relationship     = ignoreLoadedGuard( function() {
-			return invoke( this, relationshipName );
-		} );
+		var relationship = invokeRelationshipIgnoringLoadedGuard( arguments.name );
 		if ( !isObject( relationship ) || !structKeyExists( relationship, "relationshipClass" ) ) {
 			throwRelationshipNotFound( arguments.name );
 		}
@@ -2995,14 +2992,10 @@ component accessors="true" {
 		}
 
 		if ( !isRelationshipLoaded( relationshipName ) && !isLoaded() ) {
-			var relationshipArguments = arguments.missingMethodArguments;
-			var unloadedRelationship  = ignoreLoadedGuard( function() {
-				return invoke(
-					this,
-					relationshipName,
-					relationshipArguments
-				);
-			} );
+			var unloadedRelationship = invokeRelationshipIgnoringLoadedGuard(
+				relationshipName,
+				arguments.missingMethodArguments
+			);
 			unloadedRelationship.setRelationMethodName( relationshipName );
 			unloadedRelationship.initRelation( [ this ], relationshipName );
 			return retrieveRelationship( relationshipName );
@@ -3024,6 +3017,25 @@ component accessors="true" {
 		}
 
 		return retrieveRelationship( relationshipName );
+	}
+
+	/**
+	 * Invokes a relationship definition without applying loaded-entity guards.
+	 */
+	private any function invokeRelationshipIgnoringLoadedGuard(
+		required string relationshipName,
+		struct invokeArguments = {}
+	) {
+		variables._ignoreNotLoadedGuard = true;
+		try {
+			return invoke(
+				this,
+				arguments.relationshipName,
+				arguments.invokeArguments
+			);
+		} finally {
+			variables._ignoreNotLoadedGuard = false;
+		}
 	}
 
 	/**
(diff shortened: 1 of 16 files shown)


elpete:
Issue #131 is the earlier version of #177 and is covered by this implementation’s default behavior. The public regression calls to-one and to-many getters on brand-new entities, verifies null/empty-array results, and verifies no SQL runs. I added `Closes #131` and the post-fill use case to the PR review.

elpete:
## Additional issue reviews: #148 and #131

Addresses #148. **Fit score: 8/10.** `retrieveRelationship()` should distinguish missing relationship definitions from unloaded relationship values and honor defaults. This branch adds those semantics and public relationship-loading tests.

Addresses #131. **Fit score: 7/10.** Returning the relationship default for a new entity makes fill/default workflows practical, but globally returning null can hide misspelled relationship names. This implementation keeps helpful errors for nonexistent relationships while returning the defined empty/default value for valid unloaded relationships.

Issue #145, opened 2021-03-02
matchOne missing on belongsToThrough


DiscountDarcy:
If you try to eager load a belongsToThrough relationship, Quick throws this error:

```
 No matching function [MATCHONE] found
```

on BelongsToThrough.cfc:281 from BaseEntity.cfc:2475.

This function only exists on HasOneOrMany in 4.2.3.

elpete:
This relationship shouldn't need to call `matchOne` but should more closely resemble `belongsTo#match`.

elpete:
belongsToThrough eager matching is implemented and verified by the current owning-entity integration test.

Verified on `next` at [2226123](https://github.com/coldbox-modules/quick/commit/2226123f346e74e1d0f31c4db4e4806c0c5c7d9d) with qb 14.0.1. Regression: `can eager load the owning entity through other relationships` in [the integration spec](https://github.com/coldbox-modules/quick/blob/2226123f346e74e1d0f31c4db4e4806c0c5c7d9d/tests/specs/integration/BaseEntity/Relationships/BelongsToThroughSpec.cfc). The full Lucee 6 suite passed 652 tests in each null-support mode (0 failures/errors; 4 existing skips).

Closing as addressed on `next`; this does not imply a new Quick package has been released.

PR #307, opened 2026-08-22
fix: eager load one-through relationships

### LINKED ISSUE #170: `matchOne` is not defined in {BelongsToThrough, HasOneThrough}
seems `matchOne` isn't defined at the following use sites:

https://github.com/coldbox-modules/quick/blob/0c4bbcf3882df043f49a77a5fcc39ebe29611365/models/Relationships/BelongsToThrough.cfc#L281

https://github.com/coldbox-modules/quick/blob/0c4bbcf3882df043f49a77a5fcc39ebe29611365/models/Relationships/HasOneThrough.cfc#L76

elpete:
Closes #170

## Issue review

The report was correct and remained partially unresolved. `BelongsToThrough.match()` was replaced with an inline implementation in 2024, but `HasOneThrough.match()` still called an undefined inherited `matchOne()` method. This is a clear correctness bug and should be fixed.

While reproducing it through the public eager-loading API, I found the prior `BelongsToThrough` repair was also incomplete: its eager path read the wrong keys from the parent and applied joins, projections, and constraints to the related entity instead of the relationship builder that is actually executed.

## Reproduction

Tests were added first through normal Quick entity flows:
- `Country.with( "latestPost" )` reproduced `No matching function [MATCHONE] found`
- `Post.with( "country" )` first reproduced an array-key length error, then exposed that the eager query contained no joins, constraints, or `__QuickThroughKey__` projection

Before implementation, the two focused bundles reported 2 passed and 2 errors.

## Implementation

- add the missing single-result matcher to `HasOneOrManyThrough` for `HasOneThrough`
- match through results using the closest relationship’s configured local keys
- constrain `BelongsToThrough` from the parent’s foreign-key values
- select and filter using the intermediate related keys
- apply eager joins and constraints to the executed relationship builder
- retain existing lazy-loading behavior

## Validation

- focused one-through coverage: 4 passed, 0 failed, 0 errors
- adjacent through/eager/as-query coverage: 63 passed, 0 failed, 0 errors
- full suite: 497 passed, 0 failed, 0 errors, 3 skipped
- `box run-script format`
- `git diff --check`
- dependency baseline: qb `14.0.0-beta.3` (`qb@be`)

### DIFF
--- a/models/Relationships/BelongsToThrough.cfc
+++ b/models/Relationships/BelongsToThrough.cfc
@@ -108,7 +108,7 @@ component extends="quick.models.Relationships.BaseRelationship" {
 	public boolean function addEagerConstraints( required array entities, required any baseEntity ) {
 		var allKeys = getKeys(
 			entities,
-			getLocalKeys(),
+			variables.closestToParent.getForeignKeys(),
 			arguments.baseEntity
 		);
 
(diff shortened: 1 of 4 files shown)


elpete:
## Additional issue review: #145

Addresses #145. **Fit score: 10/10.** This is the same missing one-through eager-load matching behavior covered by this PR. BelongsToThrough should match like BelongsTo rather than call a nonexistent `matchOne`. The branch adds public eager-loading coverage and implements the correct matching path; there is no intentional compatibility tradeoff.

PR #308, opened 2026-08-22
fix: preserve eager loading key types

### LINKED ISSUE #166: Eager loading a belongsToMany relationship fails on Postgres
I came across several issues while working through the [Up and Running with Quick](https://github.com/Ortus-Solutions/up-and-running-with-quick) workshop and have documented those issues [here](https://github.com/Ortus-Solutions/up-and-running-with-quick/issues/2).  Most of them have a workaround that can be applied in your project, but I did run into an issue that doesn't seem solvable outside of a framework change.  

The issue occurs whenever you try to eager load a belongsToMany relationship.  This is part of Step 9 of the workshop where we add `tags` to the `with` array when initially loading  posts.

From `/handlers/Posts.cfc`
```
prc.posts = getInstance( "Post" )
	.with( [ "author", "tags" ] )
	.latest()
	.get();
```

The error generated is:

```ERROR: operator does not exist: integer = character varying Hint: No operator matches the given name and argument types. You might need to add explicit type casts. Position: 184```

💡 _Note: In my experience this error came up very frequently when working with Postgres, but required a variety of solutions.  See [the issue](https://github.com/Ortus-Solutions/up-and-running-with-quick/issues/2) I referenced above for details.  The issue I'm reporting here is for only one specific cause of this error._

In this case the problem seems to occur when calling `getKeys()` from `BelongsToMany.addEagerConstraints()`.  The **addEagerConstraints** method is what updates the WHERE clause with the needed constraints to load all tags related to all loaded posts.

Here you can see a dump of the SQL before and after the function processes the `getKeys()` result:
image
You can see that all the bindings for **postId** have **cfsqltype** set to "CF_SQL_VARCHAR".  This will trigger qb to quote the values which Postgres doesn't like (hence the error above).

I think this issue is possibly being self-inflicted in `BaseRelationship.getKeys()` when attempting to generate a unique array of primary key values arrays:
image
image

Fortunately, I think this can be fixed simply be removing the `toList()` and corresponding `map( listToArray() )`.  I confirmed that the `unique()` method would handle de-duplication of an array of arrays as well as it does an array of strings.  I will submit a PR for these changes next along with corresponding test case.

I'm still curious as to when and where this behavior was introduced in Postgres that has made it so much more pickier than MySQL.  FWIW, I've tested this in PostgreSQL versions 13.4, 12.9 and 11.0 and had the same behavior in all cases.  I'm not very familiar with Postgres at all, so If there's a flag or configuration that adjusts this strictness behavior I'm not aware of it.

elpete:
Closes #166

## Issue review

**Recommendation: 10/10.** This is a database-portability correctness bug in a core eager-loading path. Quick should preserve key value types while constructing relationship constraints, especially because PostgreSQL correctly rejects comparisons between integer columns and varchar parameters.

Reasons for:
- The current conversion through a comma-delimited list changes numeric CFML values into strings.
- The resulting `VARCHAR` bindings fail against strictly typed databases even though the entity keys were numeric.
- The fix removes a lossy internal conversion and keeps the existing public API unchanged.
- The behavior benefits all eager-loaded relationship types that share `getKeys()`.

Tradeoffs:
- Deduplication now operates on arrays of native values rather than serialized list strings. The existing relationship and composite-key suite verifies that behavior on the supported runtime.

## Reproduction

I reproduced the underlying failure through the public Quick flow `getInstance( "Post" ).with( "tags" ).get()`. The query itself succeeds on the MySQL test datasource, but the intercepted belongs-to-many eager query showed every numeric post key bound as `VARCHAR`. The new test expected `INTEGER`; before implementation the focused bundle had 27 passes and 1 failure with `Expected [INTEGER] but received [VARCHAR]`.

The repository previously received PR #167 for this issue, but it was closed without merging. This implementation revalidates the report against current `next` and current qb rather than relying on that old patch.

## Implementation

- Preserve each eager key tuple as an array of its original CFML values.
- Remove the list serialization and deserialization that coerced numeric keys to strings.
- Add a public integration regression that eager loads a belongs-to-many relationship and verifies the generated relationship bindings remain integer typed.

## Validation

- Focused EagerLoadingSpec: 28 passed, 0 failed, 0 errors
- Full suite: 496 passed, 0 failed, 0 errors, 3 skipped
- `box run-script format`
- `git diff --check`

### DIFF
--- a/models/Relationships/BaseRelationship.cfc
+++ b/models/Relationships/BaseRelationship.cfc
@@ -369,12 +369,10 @@ component accessors="true" implements="IRelationship" {
 					}
 					keyValues.append( value );
 				}
-				acc.append( keyValues.toList() );
+				acc.append( keyValues );
 				return acc;
 			}, [] )
-		).map( function( key ) {
-			return key.listToArray();
-		} );
+		);
 	}
 
 	/**
--- a/tests/specs/integration/BaseEntity/Relationships/EagerLoadingSpec.cfc
+++ b/tests/specs/integration/BaseEntity/Relationships/EagerLoadingSpec.cfc
@@ -243,6 +243,15 @@ component extends="tests.resources.ModuleIntegrationSpec" {
 				expect( variables.queries ).toHaveLength( 2, "Only two queries should have been executed." );
 			} );
 
+			it( "preserves numeric key types when eager loading a belongs to many relationship", function() {
+				getInstance( "Post" ).with( "tags" ).get();
+
+				expect( variables.queries ).toHaveLength( 2, "Only two queries should have been executed." );
+				variables.queries[ 2 ].bindings.each( function( binding ) {
+					expect( binding.cfsqltype ).toBe( "INTEGER" );
+				} );
+			} );
+
 			it( "can eager load a has many through relationship", function() {
 				var countries = getInstance( "Country" ).with( "posts" ).get();
 				expect( countries ).toBeArray();


elpete:
Closed in favor of #274

PR #312, opened 2026-08-22
test: support full-null relationship coverage

### LINKED ISSUE #154: empty belongsTo relation failing with full-null support in Lucee
When we have an empty belongsTo relation in a quick entity it fails with full null support.
The belongsTo relation is calling  the isBuilder function in qb.
The isBuilder function in QueryUtils of qb is receiving:
- an empty string with partial null support
- a null value with full null support
So this is part of the code
```return !isSimpleValue( arguments.value ) &&
        !isArray( arguments.value ) &&
        structKeyExists( arguments.value, "isBuilder" );
 ```

which will never work with null values because it will fail on the structKeyExists. I tried adding a !IsNull( arguments.value ) check but still failing, in other places I guess.

elpete:
Closes #154

## Issue review

**Recommendation: 10/10.** Empty relationships must behave correctly with full null support on every supported CFML engine. Null mode is an engine-level compatibility setting, and Quick should preserve the same relationship semantics in either representation.

Reasons for:
- A missing foreign key is normal relationship state, not an exceptional condition.
- Quick and qb should accept a real null anywhere a partial-null runtime previously supplied an empty string.
- Full-null coverage belongs in the regular CI matrix so future regressions cannot silently return.

Tradeoffs:
- Full-null and partial-null modes intentionally expose different raw values. Tests should assert public semantics where possible and narrowly normalize representation only when snapshots are testing another concern.
- Doubling the stable matrix increases CI runtime, but it validates a materially different runtime mode.

## Reproduction review

On Lucee 6 with full null support and qb `14.0.0-beta.3`, the original public flow `getInstance( "Post" ).find( 7777 ).getAuthor()` returns null without throwing. The original Quick/qb exception is fixed upstream.

The expanded full-null suite then exposed missing-key assumptions in Quick and several dependencies. Each dependency fix is maintained as a narrow CI patch here until its upstream PR is merged. Every upstream PR includes at least one public-API regression that fails before its fix.

## Implementation

- Reads `FULL_NULL` in the test application and enables engine null support.
- Runs Lucee 5/6 and BoxLang CFML 1 with both ColdBox 7/8 and full-null true/false.
- Runs Adobe 2021/2023/2025 with both null modes on ColdBox 7 and standard null mode on ColdBox 8. Adobe full-null plus ColdBox 8 currently fails inside its framework dependency graph before Quick loads, so full-null Adobe coverage uses the supported ColdBox 7 pairing.
- Runs native BoxLang 1 with ColdBox 8 in both null modes. ColdBox 7 does not expose compatible native BoxLang metadata, while BoxLang CFML continues to cover ColdBox 7.
- Adds Adobe BE on ColdBox 7 and BoxLang BE on ColdBox 8 as experimental full-null coverage.
- Expands PR, scheduled, and release workflows consistently.
- Keeps every matrix row's `fullNull` value explicit, including experimental includes, to avoid GitHub expression coercion differences.
- Preserves omitted optional arguments on Lucee 5 full-null mode when qualifying columns and excluding all global scopes. Lucee 5 materializes those omitted arguments as null keys, while newer engines may omit the keys entirely.
- Normalizes the native BoxLang date fixture using date parts so it asserts the same public value despite engine-specific date-mask parsing.

## Dependency baseline

The passing matrix resolved the latest compatible packages available during this run:
- ColdBox `8.1.0+34`
- TestBox `7.0.0+19`
- cbJavaLoader `2.6.0+5`
- qb `14.0.0-beta.3` (latest `qb@be` prerelease)
- str `4.0.2`
- mementifier `3.7.0+7`
- cfcollection `3.7.1`
- cfmigrations `5.1.3`

## Upstream dependency PRs

- TestBox: https://github.com/Ortus-Solutions/TestBox/pull/200
- cbJavaLoader: https://github.com/coldbox-modules/cbjavaloader/pull/32
- ColdBox: https://github.com/ColdBox/coldbox-platform/pull/686
- Mementifier: https://github.com/coldbox-modules/mementifier/pull/46
- cbORM (discovered while validating Mementifier): https://github.com/coldbox-modules/cborm/pull/87

Each PR documents its red/green reproduction, public regression coverage, and local engine validation.

## Validation

GitHub Actions run https://github.com/coldbox-modules/quick/actions/runs/32611389233 is green: **26 of 26 jobs passed**.

Representative full-null results:
- Lucee 6: 521 passed, 0 failed, 0 errors, 3 skipped
- Lucee 5: 521 passed, 0 failed, 0 errors, 3 skipped
- Adobe 2021 with ColdBox 7: 520 passed, 0 failed, 0 errors, 4 skipped
- Native BoxLang 1 with ColdBox 8: 521 passed, 0 failed, 0 errors, 3 skipped
- Experimental Adobe BE: 520 passed, 0 failed, 0 errors, 4 skipped
- Experimental BoxLang BE: 521 passed, 0 failed, 0 errors, 3 skipped

Also:
- workflow YAML parsed successfully
- dependency patches apply cleanly in every configured row
- `git diff --check`


elpete:
Attempted reproduction: I ran the existing public empty-belongs-to flow on a dedicated Lucee 6.2.8.20 server started with `-Dlucee.full.null.support=true` and `qb 14.0.0-beta.3`. `getInstance( "Post" ).find( 7777 ).getAuthor()` returned null and did not throw, so the original exception no longer reproduces. I then made the surrounding tests representation-independent and verified the entire suite in both full-null and normal modes: each finished with 495 passed, 0 failed, 0 errors, and 3 skipped.

PR #315, opened 2026-08-22
test: preserve bindings while chunking scoped queries

### LINKED ISSUE #151: .chunk() interferes with bindings because of .count()
We have a query that worked just fine last year on QB 7 and which is correct if we don't chunk it, but the minute we do chunk it, boolean bindings get swapped with string/GUID bindings, a BETWEEN date where clause gets a GUID instead of the two dates - never seen anything like it (on QB 8.4.8 but happens on 8.0 as well)

We can't test 7.x due to the Quick dependency - it starts out as a Quick query and then drops to QB with .retrieveQuery() but it looks like the breaking change in QB 8 was about callbacks. I don't think Quick has anything to do with this.

I can call toSQL() on the query and it's exactly right, but if I chunk it and call toSQL() on the callback, the bindings are all off

It's this line that is the culprit in chunk(): 

`var count = count( options = arguments.options );`

It works if I clone the query first by changing it to

`var count = this.clone().count( options = arguments.options );`

It seems to me (without the facility right now to make a proper repro case) that `count()` should probably always `clone()` the source query if functions like `chunk()` expect to be able to call it without mutating the query; or else something is happening in the aggregate functions to interfere with the bindings.

elpete:
## Issue review

Issue #151 reports `chunk()` mutating/reordering bindings when its internal count runs against a Quick query containing global-scope subselect bindings.

I rate the behavior **10/10** to protect. Binding order must remain stable across aggregate and pagination helpers; swapped date, boolean, string, or UUID values can silently return incorrect data or produce database errors.

## Attempted reproduction

I recreated the historical query shape through public APIs on current `next` with `qb 14.0.0-beta.3`:

- a global scope adds a subselect containing a UUID binding;
- the outer query adds a two-binding date range and a string constraint;
- the query is handed from Quick to qb with `retrieveQuery()`;
- `chunk( 1, callback )` runs its internal count and result query;
- the test verifies the correct row and the exact four-value binding order before and after chunking.

The bug did not reproduce. Current qb documents aggregate queries as non-mutating, and the source query retained the UUID, both dates, and string in the correct order.

The original report was SQL Server-specific. This repository's integration environment is MySQL, so the SQL Server execution path could not be exercised here. The regression still covers the reported mutation mechanism independently of grammar.

## Implementation

No production change was justified without a current failure. This PR strengthens the global-scope fixture with a bound subselect and adds the exact chunk/binding-order regression.

## Validation

- Focused global-scope suite: 9 passed, 0 failed, 0 errors, 1 skipped.
- Full suite: 496 passed, 0 failed, 0 errors, 3 skipped.
- `box run-script format`
- `git diff --check`

Closes #151

### DIFF
--- a/tests/resources/app/models/UserWithGlobalScope.cfc
+++ b/tests/resources/app/models/UserWithGlobalScope.cfc
@@ -8,8 +8,18 @@ component extends="User" table="users"  accessors="true" {
         qb.addSubselect( "teamName", "team.name" );
     }        
 
+    function scopeWithBoundCountryName( qb ) {
+        qb.addSubselect( "boundCountryName", function( q ) {
+            q.select( "name" )
+                .from( "countries" )
+                .whereColumn( "countries.id", "users.country_id" )
+                .where( "countries.id", "02B84D66-0AA0-F7FB-1F71AFC954843861" );
+        } );
+    }
+
     function applyGlobalScopes( qb ) {
         qb.withCountryName();
-        qb.withTeamName();        
+        qb.withTeamName();
+        qb.withBoundCountryName();
     }
 }
--- a/tests/specs/integration/BaseEntity/GlobalScopeSpec.cfc
+++ b/tests/specs/integration/BaseEntity/GlobalScopeSpec.cfc
@@ -65,6 +65,32 @@ component extends="tests.resources.ModuleIntegrationSpec" {
 				expect( user.countryName ).toBe( "United States" );
 			} );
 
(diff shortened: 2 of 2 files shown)


elpete:
Attempted reproduction note: I could not reproduce the binding corruption on current `next` with `qb 14.0.0-beta.3`. The regression uses a bound global-scope subselect followed by two date bindings and a string binding, hands the query to qb, executes `chunk()`, and verifies both the returned row and exact binding order before/after. The focused and full suites remain clean. The available integration datasource is MySQL, so I could not execute the SQL Server-specific grammar path from the original report.

PR #320, opened 2026-08-22
Test raw expression table sources

### LINKED ISSUE #141: Add support to use raw expressions in table names

elpete:
Closes #141

## Issue review

An entity-level raw table source is a good fit for Quick when applications need table hints or other qb-supported source expressions while retaining Quick attribute mapping.

Recommendation: **9/10 — implement.**

### Reasons for

- the raw source is declared once on the entity instead of repeated on every query
- qb already supports expression table sources
- Quick can keep the configured table name for column qualification while using a separate source for SQL generation
- the default remains fully backward-compatible

### Tradeoffs

- Quick cannot infer a qualifier from arbitrary SQL, so the entity's configured table name or alias remains authoritative for mapped columns
- raw source syntax is database-specific
- a custom source must expose the columns mapped by the entity

## Implementation

Entities can override `tableSource( builder )` and return any qb-supported table source, including `builder.raw(...)`. Quick uses that hook for normal entity queries, refreshed entities, reset builders, and CBORM-compatible criteria. The default implementation returns `tableName()`.

The regression entity declares `users FORCE INDEX (PRIMARY)` as its source. The test queries through the normal entity API, verifies the raw expression appears in generated SQL, and confirms the result hydrates as an entity. It does not manually replace `from()` in the test.

## Test-first evidence

The revised test was committed before the implementation and failed because the normal entity query emitted `FROM users`; the declared raw source was ignored. After adding the entity hook, the focused query suite passes 6/0/0.

## Validation

- focused QuerySpec: 6 passed, 0 failed, 0 errors
- full Lucee 6 suite: 537 passed, 0 failed, 0 errors, 3 skipped
- formatter completed
- `git diff --check` passed
- qb 14.0.0-beta.3

### DIFF
--- a/models/BaseEntity.cfc
+++ b/models/BaseEntity.cfc
@@ -363,6 +363,21 @@ component accessors="true" {
 	}
 
 	/**
+	 * Returns the table source for this entity's queries.
+	 *
+	 * Override this method to return a raw expression or another qb-supported
+	 * table source. The configured table name remains available for qualifying
+	 * entity columns.
+	 *
+	 * @builder  The builder that will receive the table source.
+	 *
+	 * @return   any
+	 */
+	public any function tableSource( required any builder ) {
+		return tableName();
+	}
+
+	/**
 	 * Returns the table name for this entity.
 	 *
 	 * @return  String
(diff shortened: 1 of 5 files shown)


elpete:
Attempted the original raw-table reproduction against current `next` and qb 14.0.0-beta.3 using Quick’s public `newQuery().from(raw(...)).where(...).firstOrFail()` flow. The historical qualification error no longer reproduces: the query executes and hydrates the expected entity. The PR adds that exact regression guard; the implementation history and remaining alias limitation are documented in the PR body.

elpete:
Closing because issue #141 is covered by Quick’s existing entity global-scope hook and qb’s raw source API.

PR #331, opened 2026-08-22
feat: create belongs-to-many related entities

### LINKED ISSUE #84: .create() support on belongsToMany and hasManyThrough
`.attach()` will create an intermediate entity (with one caveat) but there is no analog for `.create()` on entities involving pivot tables. The function works -- it creates the target entity -- but it doesn't create the intermediate entity, and it returns the relationship object rather than the newly created entity.

Two things to solve:

1) Implement .create() on these relationships in such a way that the intermediate entity is also created, and
2) Either supply a mechanism to populate non-key fields on the intermediate entity, or else require that the intermediate entity be a Quick entity, so we can use interception points like `preSave` if we need fields set besides the two key fields.

elpete:
Closes #84

## Review

This is the second PR in the belongs-to-many stack and now contains only relationship `create()` support. Pivot models, pivot attributes, constraints, timestamps, and persistence are implemented separately in #353.

Both PRs target `next`. Until #353 merges, GitHub will show its commits in this PR as the shared foundation. After #353 lands, this PR's production diff is one `BelongsToMany.create()` method.

The two relationship types requested by the issue have different semantics:

- **`belongsToMany.create`: 9/10.** The parent, related entity, and one pivot row are unambiguous.
- **Implicit `hasManyThrough.create`: 3/10.** A parent can have several intermediate entities or an arbitrarily deep chain, so Quick cannot infer which intermediate row should own the terminal entity.

This PR implements only the unambiguous belongs-to-many case.

## Implementation

```cfml
var tag = post.tags().create(
    { name : "testing" },
    { context : "created through relationship" }
);
```

The method:

1. Creates and returns the loaded related entity.
2. Attaches that entity to the parent.
3. Passes the optional pivot attributes to the pivot support from #353.

The related insert and pivot insert remain separate operations. Quick does not imply aggregate or unit-of-work persistence; callers requiring atomicity should provide their transaction boundary.

## Test-first evidence

Before implementation, the public `post.tagsWithPivot().create(...)` call errored because `create` was forwarded to qb and did not exist. The test now verifies the returned entity and reloads the relationship through Quick's public API to verify both the association and pivot values.

## Validation

- Focused stacked belongs-to-many suite: 12 passed, 0 failed, 0 errors
- Full stacked suite on Lucee 6: 567 passed, 0 failed, 0 errors, 3 skipped
- Formatter completed
- `git diff --check` passed

Depends on #353.

### DIFF
--- a/models/Relationships/BelongsToMany.cfc
+++ b/models/Relationships/BelongsToMany.cfc
@@ -650,6 +650,31 @@ component
 	}
 
 	/**
+	 * Creates a new related entity and attaches it to the parent through the pivot table.
+	 *
+	 * @attributes                   Attributes for the related entity.
+	 * @pivotAttributes              Additional attributes for the pivot row.
+	 * @ignoreNonExistentAttributes  Whether to ignore attributes not defined on the related entity.
+	 * @options                      Options passed to the related entity save query.
+	 *
+	 * @return  quick.models.BaseEntity
+	 */
+	public any function create(
+		struct attributes                   = {},
+		struct pivotAttributes              = {},
+		boolean ignoreNonExistentAttributes = false,
+		struct options                      = {}
+	) {
+		var entity = variables.related.create(
+			arguments.attributes,
+			arguments.ignoreNonExistentAttributes,
+			arguments.options
+		);
+		attach( entity, arguments.pivotAttributes );
+		return entity;
+	}
+
+	/**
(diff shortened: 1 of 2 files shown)


elpete:
Final validation after restacking on #353: all 27 CI checks are green. Once #353 merges, this PR reduces to the public-API regression test and the BelongsToMany.create() implementation.

PR #333, opened 2026-08-22
Test Quick with Lucee full null support

### LINKED ISSUE #64: Test with full null support

elpete:
Closes #64

## Issue review

**Recommendation: 10/10 — implement.** Full null support changes how database nulls are represented at Quick’s public boundaries, so exercising this mode in CI is a strong fit. It catches compatibility regressions that the default Lucee configuration necessarily hides. The tradeoff is one additional Lucee job per ColdBox matrix entry and null-mode-aware assertions where the two engine modes intentionally differ.

## Implementation

- adds a dedicated Lucee 6 server profile with `lucee.full.null.support=true`
- adds that profile to PR, cron, and release test matrices
- verifies database null behavior through normal Quick entity getters, mementos, casts, relationships, saves, and qb query results
- centralizes full-null detection for integration specs
- keeps memento null-default behavior distinct from raw entity/query null behavior

## Test-first evidence

The first full-null baseline reproduced seven failing assertions (and zero errors) where tests assumed Lucee would coerce database nulls to empty strings. The assertions were updated to validate each mode’s intentional public behavior.

## Validation

- Lucee 6 full-null suite: 495 passed, 0 failed, 0 errors, 3 skipped
- Lucee 6 standard suite: 495 passed, 0 failed, 0 errors, 3 skipped
- `box run-script format`
- `git diff --check`

Uses `qb@14.0.0-beta.3`.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        cfengine: ["lucee@5", "lucee@6", "adobe@2021", "adobe@2023", "adobe@2025", "boxlang-cfml@1"]
+        cfengine: ["lucee@5", "lucee@6", "lucee@6-full-null", "adobe@2021", "adobe@2023", "adobe@2025", "boxlang-cfml@1"]
         coldbox: ["coldbox@^7", "coldbox@^8"]
         experimental: [ false ]
         include:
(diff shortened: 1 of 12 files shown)


elpete:
Superseded by merged PR #312. That PR now runs the broader full-null matrix across Lucee, Adobe CF, native BoxLang, and BoxLang CFML, and includes null-safe public-flow coverage. Rebasing this older branch would reintroduce a redundant Lucee-only server profile and duplicate assertions, so there is no remaining independent change to review.

PR #339, opened 2026-08-22
Add optional parallel eager loading

### LINKED ISSUE #55: Optional async eager loading
Each `with` relationship should be able to be parallelized. Add an optional argument or setting to do this.

elpete:
Closes #55

## Issue review

**Recommendation: 7/10 — implement as opt-in.** Independent eager-load branches can benefit from concurrency when database latency and hydration work dominate, especially for several unrelated top-level relationships. It should not be the default: concurrency consumes more database connections and CPU, and does not help a single relationship or necessarily improve low-latency queries.

## Implementation

- adds an optional `parallel` boolean to `with`, e.g. `.with( [ "author", "comments" ], true )`
- uses ColdBox's AsyncManager and an application-wide bounded executor; applications can supply one with `parallelEagerLoadingExecutor`, otherwise Quick creates and owns a fixed executor
- caps each query's `parallelEagerLoadingMaxThreads` by the configured executor's maximum pool size
- performs retrieval, entity hydration, transformations, nested eager loading, and branch-local collection construction on workers
- keeps matching and mutation of the original parent entities, plus top-level `quickRelationshipLoaded`, on the calling thread
- hydrates polymorphic branches on workers without allowing them to mutate the original parents
- gives every worker an isolated ColdBox request context and preserves the BoxLang application mappings and default datasource needed during hydration
- propagates worker failures instead of returning partial relationship data, cancels outstanding work on failure or timeout, and avoids parallel work inside transactions
- retains the existing sequential path by default and for a single top-level relationship

## Validation

- focused relationship/module suite on Lucee 5: 59 passed, 0 failed, 0 errors
- focused relationship/module suite on BoxLang 1: 59 passed, 0 failed, 0 errors
- full Lucee 5 suite: 624 passed, 0 failed, 0 errors, 3 skipped
- full BoxLang 1 suite: 625 passed, 0 failed, 0 errors, 2 skipped
- `git diff --check`

Uses `qb@14.0.0-beta.3`.


elpete:
Integrated into next with the current automatic-timestamp state and entity-definition cleanup retained. Local validation against published qb 14.0.1: Lucee 6 full suite passed 674 tests in both null modes (4 existing skips); BoxLang 1 full suite passed 676 tests (2 existing skips). Focused eager-loading coverage passed 57 tests on both engines. CI for the integration commit is the remaining multi-engine verification.

